PR #3, opened 2019-12-13
Create terraform environment for frontend

MancunianSam:


modules/transfer-frontend/route53.tf
```diff
@@ -0,0 +1,26 @@
+//resource "aws_route53_zone" "main" {
```
TomJKing: Should remove commented out code

modules/transfer-frontend/vpc.tf
```diff
@@ -0,0 +1,100 @@
...
+  )
+}
+
+# Explicitly associate the newly created route tables to the private subnets (so they don't default to the main route table)
+resource "aws_route_table_association" "private" {
+  count          = var.az_count
+  subnet_id      = aws_subnet.private.*.id[count.index]
+  route_table_id = aws_route_table.private.*.id[count.index]
+}
+
```
TomJKing: remove unnecessary line breaks

PR #22, opened 2020-01-22
Create migration lambda

MancunianSam:


modules/consignment-api/database.tf
```diff
@@ -0,0 +1,59 @@
+resource "random_password" "password" {
```
suzannehamilton: Just wondering - do you know how we can do password rotation without API downtime? Might be worth adding to the tech backlog.

modules/consignment-api/database.tf
```diff
@@ -0,0 +1,59 @@
...
+  length  = 16
```
suzannehamilton: Probably worth increasing this, given that nobody's (hopefully) ever going to have to type it.

modules/consignment-api/database.tf
```diff
@@ -0,0 +1,59 @@
...
+
+resource "random_string" "snapshot_prefix" {
+  length  = 4
+  upper   = false
+  special = false
+}
+
+resource "aws_rds_cluster" "consignment_api_database" {
+  cluster_identifier_prefix = "consignment-api-db-${var.environment}"
+  engine                    = "aurora-mysql"
```
suzannehamilton: I thought this was going to be Postgres like in the prototype. Did that decision change?

modules/consignment-api/ecs.tf
```diff
@@ -0,0 +1,127 @@
...
+    var.common_tags,
+    map("Name", "consignment_api_${var.environment}")
+  )
+}
+
+data "template_file" "app" {
+  template = file("modules/consignment-api/templates/consignment-api.json.tpl")
+
+  vars = {
+    app_image       = "nationalarchives/sangria:${var.environment}"
```
suzannehamilton: Should the image also be called something like "consignment-api" for consistency?

modules/consignment-api/outputs.tf
```diff
@@ -0,0 +1,11 @@
+output "database_url" {
+  value = aws_rds_cluster.consignment_api_database.endpoint
+}
+
+output "database_username" {
+  value = aws_rds_cluster.consignment_api_database.master_username
+}
+
+output "database_password" {
+  value = aws_rds_cluster.consignment_api_database.master_password
```
suzannehamilton: I don't know if this applies to modules, but it looks like you can mark an output as sensitive so it's not printed in the console: https://www.terraform.io/docs/configuration/outputs.html#sensitive-suppressing-values-in-cli-output

modules/database-migrations/lambda.tf
```diff
@@ -0,0 +1,93 @@
...
+    ]
+    resources = [
+      "arn:aws:s3:::tdr-database-migrations/*",
+      "arn:aws:s3:::tdr-database-migrations"
+    ]
+  }
+  statement {
+    effect = "Allow"
+    actions = [
+      "ec2:DeleteNetworkInterface",
```
suzannehamilton: Might be a silly question, but why does the lambda need these permissions? Is it because it's running in a private subnet?

modules/database-migrations/lambda.tf
```diff
@@ -0,0 +1,93 @@
...
+}
+
+resource "aws_lambda_function" "database_migration_function" {
+  function_name = "tdr-database-migrations-${var.environment}"
+  handler       = "migration.Main::runMigration"
+  role          = aws_iam_role.lambda_assume_role.arn
+  runtime       = "java8"
+  filename      = "${path.module}/temp.zip"
+  memory_size   = 128
+  timeout       = 60
```
suzannehamilton: Should this be higher, just in case we have a DB migration that takes a while in production?

What do you think about having some kind of warning if we run a migration that runs close to the limit? e.g. the Jenkins migration job could be marked as unstable if it takes too long.

root.tf
```diff
@@ -27,22 +28,56 @@ provider "aws" {
...
   region                      = "eu-west-2"
+  db_migration_sg             = module.database_migrations.db_migration_security_group
+}
+
+module "frontend" {
+  app_name        = "frontend"
+  source          = "./modules/transfer-frontend"
+  environment     = local.environment
+  common_tags     = local.common_tags
+  region          = "eu-west-2"
```
suzannehamilton: Minor thing, but the region should probably be a variable so we only define it in one place.

modules/consignment-api/database.tf
MancunianSam: I've added it to the github project and put a comment on the card with what you suggested about creating a second user.

modules/consignment-api/database.tf
MancunianSam: I've put it up to 64. I can't find anything about password length limits for rds but this seems sensible.

modules/consignment-api/database.tf
MancunianSam: We were going to use serverless but that didn't work with Flyway's free version so rather than changing the code back to postgres, I changed it to non-serverless mysql. We can always change it later.

modules/consignment-api/ecs.tf
MancunianSam: It should, I've changed it.

modules/consignment-api/outputs.tf
MancunianSam: Not sure if it applies. I've added it in anyway, it can't hurt.

modules/database-migrations/lambda.tf
MancunianSam: Yeah, if it's inside a VPC, it needs these.

modules/database-migrations/lambda.tf
MancunianSam: I think this is ok for now and for a while yet. Anything running close to this at the moment is probably due to a failure which would be useful to know sooner.

root.tf
MancunianSam: I've swapped them out for a local variable.

modules/consignment-api/database.tf
MancunianSam: Apparently the max length is 41, I've set it to that.

modules/database-migrations/lambda.tf
suzannehamilton: I've [created a card](https://github.com/orgs/nationalarchives/projects/20#card-32043820) for adding some kind of warning, and put it low down on the tech backlog.

PR #24, opened 2020-01-31
Tk update readme

TomJKing:


README.md
```diff
@@ -72,7 +72,9 @@ HCL Language Support: https://plugins.jetbrains.com/plugin/7808-hashicorp-terraf
    source_profile = terraform   
    ```
    
-## Running Terraform Project
+## Running Terraform Project Locally
+
+**NOTE: Running Terraform locally should only be used to check the TErraform plan. Updating the TDr environments should only ever be done through Jenkins**
```
suzannehamilton: Creative capitalization: TErraform and TDr

README.md
```diff
@@ -72,7 +72,9 @@ HCL Language Support: https://plugins.jetbrains.com/plugin/7808-hashicorp-terraf
    source_profile = terraform
```
suzannehamilton: Did you say that these profiles aren't relevant any more?

README.md
```diff
@@ -101,13 +103,19 @@ HCL Language Support: https://plugins.jetbrains.com/plugin/7808-hashicorp-terraf
    
    There is an issue with Terraform not using the correct profile when more than one profile in the AWs config file
    
-6. Initialize Terraform (if not done so previously):
+6. Set the following Terraform environment variables on the local environment:
+
+    * TF_VAR_tdr_account_number=*[account number of the environment to update]*
+    * TF_VAR_tdr_environment=*[name of the TDR environment to update]*
```
suzannehamilton: Is this step necessary if we're using workspaces? I think you can have workspace-specific variables, but I might be misremembering.

TomJKing: Not sure I undestand what you mean here? https://www.terraform.io/docs/configuration/variables.html

suzannehamilton: Sorry, I misremembered how Terraform worked.

I was just wondering if there was a way to avoid setting a `TF_VAR_tdr_environment` variable, since we should be able to work out the environment from the workspace. Would it work if you used `terraform.workspace` instead of `var.tdr_environment`?

TomJKing: Added as terraform.workspace so no need to set the TF_var

PR #26, opened 2020-01-31
Add route 53

MancunianSam:


modules/consignment-api/variables.tf
```diff
@@ -6,4 +6,5 @@ variable "public_subnets" {}
 variable "private_subnets" {}
 variable "database_availability_zones" {}
 variable "region" {}
-variable "db_migration_sg" {}
\ No newline at end of file
+variable "db_migration_sg" {}
+variable "environment_full" {}
```
TomJKing: Could this be changed to environment_full_name just to make it clear what it is?

root.tf
```diff
@@ -1,7 +1,11 @@
 locals {
   environment = var.tdr_environment
   assume_role = "arn:aws:iam::${var.tdr_account_number}:role/TDRTerraformRole${title(local.environment)}"
-
+  environment_full_map = {
```
TomJKing: Could this be changed to environment_full_name_map?

modules/consignment-api/variables.tf
MancunianSam: I've changed the name here and for the map as well.

PR #27, opened 2020-01-31
VPC flowlogs

paulschwarzenberger:


modules/keycloak/flowlogs.tf
```diff
@@ -0,0 +1,66 @@
...
+    actions = ["sts:AssumeRole"]
+    principals {
+      type        = "Service"
+      identifiers = ["vpc-flow-logs.amazonaws.com"]
+    }
+  }
+}
+
+resource "aws_iam_policy" "keycloak_flowlog_policy" {
+  name   = "keycloakkeycloakFlowlogPolicy${title(var.environment)}"
```
suzannehamilton: Is "keycloakkeycloak" a typo?

modules/keycloak/flowlogs.tf
```diff
@@ -0,0 +1,66 @@
...
+    actions = [
+      "logs:CreateLogGroup",
+      "logs:CreateLogStream",
+      "logs:PutLogEvents",
+      "logs:DescribeLogGroups",
+      "logs:DescribeLogStreams"
+    ]
+
+    resources = [
+      "*",
```
suzannehamilton: Does it make sense to tie this to the specific cloudwatch log group?

Ditto for the policy below.

modules/keycloak/flowlogs.tf
suzannehamilton: Sorry, out-of-date comment.

modules/keycloak/flowlogs.tf
paulschwarzenberger: Thanks, good idea, updated

PR #28, opened 2020-02-04
Fix environment variable

TomJKing:
Get TDR environment from local instead of root variable

Point intg account to master branch in Jenkins build

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -77,7 +77,7 @@ pipeline {
 def getStageFromBranch() {
 
     def branchToStageMap = [
-            "origin/intg": "intg",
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -77,7 +77,7 @@ pipeline {
 def getStageFromBranch() {
 
     def branchToStageMap = [
-            "origin/intg": "intg",
+            "origin/master": "intg",
```
suzannehamilton: Will it eventually be switched back to intg once we've set up the new branching scheme?

TomJKing: I guess so, if that is what we will be doing with the branching

PR #33, opened 2020-02-07
KMS key for RDS database encryption

paulschwarzenberger:


modules/consignment-api/database.tf
```diff
@@ -21,6 +21,22 @@ resource "random_string" "snapshot_prefix" {
   special = false
 }
 
+resource "aws_kms_key" "encryption" {
+  description         = "KMS key for encryption within the environment"
+  enable_key_rotation = true
```
suzannehamilton: As discussed in person, I've looked into what key rotation means for RDS instances.

When the KMS key is rotated, [the DB is _not_ rencrypted](https://forums.aws.amazon.com/message.jspa?messageID=770958#770910). The data continues to be encrypted with the original value of the CMK. KMS stores all previous values of the CMK so you can keep decrypting old data.

I'm a bit confused about whether RDS uses the CMK itself, or a data key encrypted by the CMK. Most of the docs I've seen imply the former ([example](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.Encryption.html)) but [this diagram](https://aws.amazon.com/rds/features/security/#Encryption_of_Data_at_Rest) implies the latter. But I think the upshot is the same: the RDS DB is never re-encrypted, unless you do it manually by creating and restoring a snapshot using a new key.

modules/shared-resources/key.tf
```diff
@@ -0,0 +1,15 @@
+resource "aws_kms_key" "encryption" {
```
suzannehamilton: Like we chatted about in person, at some point it would be good to think about whether some encrypted data is more sensitive (e.g. the S3 buckets that contain uploaded files) and need separate keys.

modules/consignment-api/database.tf
paulschwarzenberger: Thanks for researching this and clarifying!

PR #38, opened 2020-03-02
ALB module in tdr-terraform-modules

paulschwarzenberger:


root.tf
```diff
@@ -94,10 +62,77 @@ module "frontend" {
...
   environment_full_name       = local.environment_full_name_map[local.environment]
   common_tags                 = local.common_tags
   database_availability_zones = local.database_availability_zones
   az_count                    = 2
   region                      = local.region
 }
+
+module "alb_logs_s3" {
+  source        = "./tdr-terraform-modules/s3"
+  project       = "tdr"
```
suzannehamilton: Should this be `var.project` like the others?

root_variables.tf
```diff
@@ -2,3 +2,18 @@ variable "tdr_account_number" {
   description = "The AWS account number where the TDR environment is hosted"
   type        = string
 }
+
+variable "dns_zone" {
+  description = "DNS zone name e.g. tdr-management.nationalarchives.gov.uk"
+  default = "tdr-management.nationalarchives.gov.uk"
+}
+
+variable "function" {
```
suzannehamilton: I might just be missing something, but are `var.dns_zone` and `var.function` used anywhere?

README.md
```diff
@@ -54,7 +54,12 @@ HCL Language Support: https://plugins.jetbrains.com/plugin/7808-hashicorp-terraf
 
 2. In command terminal navigate to the folder where the project has been cloned to
 
-3. Create Terraform workspaces corresponding to the TDR environments:
+3. Clone Terraform modules repository
+   ```
+   [location of project] $ git clone git@github.com:nationalarchives/tdr-terraform-environments.git
```
suzannehamilton: Should this be tdr-terraform-modules rather than tdr-terraform-environments?

As discussed in person, it would be nice if we didn't have to clone tdr-terraform-modules into every Terraform project, but we can't use Terraform variables to get round this because [Terraform doesn't allow you to use variables in the module source parameter](https://github.com/hashicorp/terraform/issues/1439).

paulschwarzenberger: Thanks, corrected

root.tf
paulschwarzenberger: Thanks, corrected

root_variables.tf
paulschwarzenberger: Thanks, removed both variables as they are not being used

PR #45, opened 2020-03-24
Add frontend url var

MancunianSam:


root_locals.tf
```diff
@@ -19,7 +19,7 @@ locals {
   )
   database_availability_zones = ["eu-west-2a", "eu-west-2b"]
 
-  region                      = "eu-west-2"
+  region = "eu-west-2"
```
TomJKing: Did you run terraform fmt?

MancunianSam: Yeah, I ran it on all files, this one wasn't formatted before.

PR #48, opened 2020-04-17
Keycloak cognito integration

MancunianSam:


modules/transfer-frontend/templates/cognito_authenticated.json.tpl
```diff
@@ -0,0 +1,42 @@
...
+      "Sid": "ListYourObjects",
+      "Effect": "Allow",
+      "Action": "s3:ListBucket",
+      "Resource": [
+        "arn:aws:s3:::tdr-upload-files-${environment}"
+      ],
+      "Condition": {
+        "StringLike": {
+          "s3:prefix": [
+            "$${cognito-identity.amazonaws.com:sub}"
```
TomJKing, line 14: Should that be a "$$"?

modules/transfer-frontend/cognito.tf
```diff
@@ -0,0 +1,61 @@
...
+
+resource "aws_cognito_identity_pool_roles_attachment" "tdr_frontend_roles_attachment" {
+  identity_pool_id = aws_cognito_identity_pool.tdr_frontend_identity_pool.id
+  roles = {
+    authenticated = aws_iam_role.cognito_authorised_role.arn
+  }
+}
+
+resource "aws_iam_role_policy_attachment" "tdr_frontend_role_attachment" {
+  policy_arn = aws_iam_policy.cognito_auth_policy.arn
```
TomJKing, line 21: Did you run fmt?

root.tf
```diff
@@ -53,6 +53,7 @@ module "frontend" {
   public_subnets        = module.shared_vpc.public_subnets
   private_subnets       = module.shared_vpc.private_subnets
   dns_zone_name_trimmed = local.dns_zone_name_trimmed
+  auth_url = module.keycloak.auth_url
```
TomJKing, line 56: Did you run fmt?

modules/transfer-frontend/templates/cognito_authenticated.json.tpl
MancunianSam: Yeah, because the string in the policy needs to be `${cognito-identity.amazonaws.com:sub}` but terraforms string interpolation uses ${} This is how you get it to just use the string you want in the policy.

modules/transfer-frontend/cognito.tf
MancunianSam: I did not.

modules/transfer-frontend/templates/cognito_authenticated.json.tpl
TomJKing: Fair enough

PR #49, opened 2020-04-22
Remove create log group permission.

MancunianSam:
This is because of an issue where the flow log cloudwatch log groups
were being recreated after terraform destroy was run. This prevented us
from running terraform apply because they already existed. This is
probably due to this github issue
https://github.com/hashicorp/terraform/issues/14750#issuecomment-303337513

The policy doesn't need to be able to create log groups because the log
group is created by terraform and removing this permission prevents it
recreating the log group after terraform destroy.


paulschwarzenberger:
Looks good to me, you may want to consider adding "logs:DescribeLogStreams" to the policy as recommended by others

MancunianSam:
logs:DescribeLogStreams is already in the policy, it's just not showing on the PR.

PR #58, opened 2020-05-05
Add keycloak backend checks secret

TomJKing:
Updating the database through the API with the backend checks data, requires authorisation via a keycloak service account, as backend checks triggerd by lambda, not a user

Keycloak service account authorisation requires additional secret to allow for retrieval of token by client id and client secret: https://www.keycloak.org/docs/latest/server_admin/index.html

Create and expose the backend checks secret to the relevant TDR modules to allow updating of the backend checks data via lambda

### DIFF
--- a/modules/keycloak/ecs.tf
+++ b/modules/keycloak/ecs.tf
@@ -23,7 +23,7 @@ data "template_file" "app" {
     password_path                     = aws_ssm_parameter.database_password.name
     admin_user_path                   = aws_ssm_parameter.keycloak_admin_user.name
     admin_password_path               = aws_ssm_parameter.keycloak_admin_password.name
-    frontend_app_client_secret_path                = aws_ssm_parameter.keycloak_frontend_app_client_secret.name
+    frontend_app_client_secret_path   = aws_ssm_parameter.keycloak_frontend_app_client_secret.name
     backend_checks_client_secret_path = aws_ssm_parameter.keycloak_backend_checks_client_secret.name
     frontend_url                      = var.frontend_url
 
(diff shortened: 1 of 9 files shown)


modules/keycloak/ecs.tf
```diff
@@ -14,17 +14,18 @@ data "template_file" "app" {
...
+    app_image                         = "nationalarchives/tdr-auth-server:${var.environment}"
+    app_port                          = local.app_port
+    app_environment                   = var.environment
+    aws_region                        = var.region
+    url_path                          = aws_ssm_parameter.database_url.name
+    username_path                     = aws_ssm_parameter.database_username.name
+    password_path                     = aws_ssm_parameter.database_password.name
+    admin_user_path                   = aws_ssm_parameter.keycloak_admin_user.name
+    admin_password_path               = aws_ssm_parameter.keycloak_admin_password.name
+    client_secret_path                = aws_ssm_parameter.keycloak_client_secret.name
```
suzannehamilton, line 26: How much disruption would it cause to rename this parameter to distinguish it from your new client secret?

TomJKing: Its going to need a lot of changes in a lot of places. I'm not sure its worth the effort.

suzannehamilton: Ah, OK. It might be worth doing it if we find it causes confusion, but let's leave it for now.

TomJKing: I've actually implemented the changes in this PR

suzannehamilton: Sorry, I misunderstood! Thanks for updating the name. 👍

PR #59, opened 2020-05-05
Antivirus lambda

MancunianSam:
This includes the antivirus lambda, the quarantine bucket and the bucket to store the lambda function code.

### DIFF
--- a/root.tf
+++ b/root.tf
@@ -224,12 +224,9 @@ module "backend_lambda_function_bucket" {
 module "av_lambda" {
   source            = "./tdr-terraform-modules/lambda"
   project = var.project
-  function = "yara-av"
   environment = local.environment
   common_tags = local.common_tags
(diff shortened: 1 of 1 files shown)


root.tf
```diff
@@ -206,3 +213,18 @@ module "waf" {
...
 }
+
+module "backend_lambda_function_bucket" {
+  source = "./tdr-terraform-modules/s3"
+  common_tags = local.common_tags
+  function = "backend-checks"
+  project = var.project
+}
+
+module "av_lambda" {
```
suzannehamilton, line 224: Would you mind renaming this from "av" to "antivirus" so it's clear to someone who's not familiar with what backend checks TDR does? I always think "audiovisual" before I remember it stands for "antivirus".

MancunianSam: Yeah sure, I've renamed it.

PR #68, opened 2020-05-14
Restrict IPs to Integration environment

paulschwarzenberger:


modules/transfer-frontend/security.tf
```diff
@@ -7,14 +7,14 @@ resource "aws_security_group" "lb" {
     protocol    = "tcp"
     from_port   = 443
     to_port     = 443
-    cidr_blocks = ["0.0.0.0/0"]
+    cidr_blocks = local.environment == "intg" ? var.ip_whitelist : ["0.0.0.0/0"]
```
suzannehamilton, line 10: Minor thing, but is the `local.environment == "intg"` necessary? Could we make the `0.0.0.0/0` the variable default?

I think it would make it a bit easier to follow what's going on. For example if we were to pass in an `ip_whitelist` variable in staging, this check means it wouldn't work. We'd have to track down this conditional to find the problem.

README.md
```diff
@@ -52,14 +52,16 @@ HCL Language Support: https://plugins.jetbrains.com/plugin/7808-hashicorp-terraf
 
 1. Clone TDR Environments project to local machine: https://github.com/nationalarchives/tdr-terraform-environments
 
-2. In command terminal navigate to the folder where the project has been cloned to
+2. Clone TDR Configurations project to local machine: https://github.com/nationalarchives/tdr-configurations
```
suzannehamilton, line 55: Thanks for catching this. 👍

modules/transfer-frontend/security.tf
paulschwarzenberger: updated as suggested, with conditional in root

paulschwarzenberger:
IPs restricted to Frontend for Integration, no changes to KeyCloak or Consignment API.

PR #69, opened 2020-05-14
Api update antivirus lambda

MancunianSam:


root.tf
```diff
@@ -263,4 +264,14 @@ module "file_format_sqs_queue" {
   function       = "file-format"
   sns_topic_arns = [module.dirty_upload_sns_topic.sns_arn]
   sqs_policy     = "sns_topic"
-}
\ No newline at end of file
+}
+
+
+module "api_update_antivirus_lambda" {
```
suzannehamilton, line 270: Is this lambda just for the antivirus, or will it be used by all the backend checks?

MancunianSam: Just fo the antivirus, I tried to get a single lambda to do all of the api updates and it's a bit of a nightmare in terms of parsing json so it's simpler to have one each in a multi project build.

modules/consignment-api/outputs.tf
```diff
@@ -14,3 +14,7 @@ output "database_password" {
   value     = aws_rds_cluster.consignment_api_database.master_password
   sensitive = true
 }
+
+output "api_url" {
+  value = "https://${aws_route53_record.api_dns.name}.${var.dns_zone_name_trimmed}"
```
suzannehamilton, line 19: Out of curiosity, how does the frontend get this URL, and should it also use this output?

root.tf
suzannehamilton: Do you want me to have a go at the parsing, if you think it's worth pursuing? Or do you want to try out having separate queues and lambdas and see how it goes?

MancunianSam: Separate queues is the way to go. It can be done, I could have got the parsing working but I think this is the cleaner way. We can always merge them later if it becomes difficult.

modules/consignment-api/outputs.tf
MancunianSam: The frontend has one application conf file per environment so the urls are all hard coded which is why we've never needed this before.

PR #72, opened 2020-06-05
Trigger end to end tests on intg deployment

TomJKing:


Jenkinsfile
```diff
@@ -75,15 +77,12 @@ pipeline {
             echo 'Deleting Jenkins workspace...'
             deleteDir()
         }
+        success {
+            script {
+                if (params.STAGE == "intg"){
+                    tdr.runEndToEndTests(300, params.STAGE, BUILD_URL)
```
suzannehamilton, line 83: I approve this earlier, but this just occurred to me: the ECS deployments need a delay before running the end-to-end tests, but maybe the Terraform job doesn't? Is there anything that would take a while to break the end-to-end tests?

TomJKing: Not 100% sure, but maybe a Terraform change that causes a new ECS task definition to be created, or something along those lines.

suzannehamilton: Shall we go with this for now, and see if the 5 minute wait causes any issues?

TomJKing: Sure

PR #74, opened 2020-06-10
Create DLQ and add to backend check queues.

MancunianSam:


root.tf
```diff
@@ -250,47 +250,64 @@ module "dirty_upload_sns_topic" {
   sns_policy  = "s3_upload"
 }
 
+module "backend_check_sqs_queue" {
```
TomJKing, line 253: Maybe rename the module: "backend_check_failure_sqs_queue"?

MancunianSam: Yeah, good point. I've changed it.

PR #75, opened 2020-06-12
Use single api update lambda

MancunianSam:


root.tf
```diff
@@ -290,31 +290,21 @@ module "file_format_sqs_queue" {
...
-  redrive_maximum_receives = 3
-}
-
-module "api_update_antivirus_lambda" {
+module "api_update_lambda" {
   source                                = "./tdr-terraform-modules/lambda"
   project                               = var.project
   common_tags                           = local.common_tags
-  lambda_api_update_av                  = true
+  lambda_api_update                  = true
```
TomJKing, line 307: Did you run terraform fmt?

MancunianSam: I didn't. I've run it now.

PR #79, opened 2020-06-30
Replace space with percent encoding

techncl:
Whenever Jenkins posts a Terraform message in Slack, the link in broken due to the spaces in the job name. This PR replaces spaces with the percent encoding

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -44,7 +44,7 @@ pipeline {
                             tdr.postToDaTdrSlackChannel(colour: "good",
                                                         message: "Terraform plan complete for ${params.STAGE} TDR environment." +
                                                                  "View here for plan: https://jenkins.tdr-management.nationalarchives.gov.uk/job/" +
-                                                                 "${JOB_NAME.replaceAll(" ", "%20")}/${BUILD_NUMBER}/console"
+                                                                 "${JOB_NAME.replaceAll(' ', '%20')}/${BUILD_NUMBER}/console"
                             )
                         }
                     }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -43,7 +43,8 @@ pipeline {
                         script {
                             tdr.postToDaTdrSlackChannel(colour: "good",
                                                         message: "Terraform plan complete for ${params.STAGE} TDR environment." +
-                                                                 "View here for plan: https://jenkins.tdr-management.nationalarchives.gov.uk/job/${JOB_NAME}/${BUILD_NUMBER}/console"
+                                                                 "View here for plan: https://jenkins.tdr-management.nationalarchives.gov.uk/job/" +
+                                                                 "${JOB_NAME.replaceAll(' ', '%20')}/${BUILD_NUMBER}/console"
```
suzannehamilton, line 47: Good spot! I don't know if Jenkins will allow this, but is it possible to call [URLEncoder](https://docs.oracle.com/javase/7/docs/api/java/net/URLEncoder.html#encode(java.lang.String,%20java.lang.String)), which will encode any special characters including spaces?

techncl: 1. I created a replaceAll method test on Jenkins yesterday and had it post to Slack and it worked. When I use the URLEncoder.encode(), it replaces the spaces with a + but Jenkins adds a "%20" in the place of spaces for its job pages,  e.g.

https://jenkins.tdr-management.nationalarchives.gov.uk/job/TDR%20AWS%20Accounts%20Deploy/78/

if I use "+" in place of the spaces, the page returns a 404

https://jenkins.tdr-management.nationalarchives.gov.uk/job/TDR+AWS+Accounts+Deploy/78/

2. Also, just noticed that for some reason, one of the two urls I changed has "/input" at the end of it. When I add "/input" at the end of a Jenkins' url, it doesn't take me anywhere, I'm not sure if that's because the page does not exist or because I don't have access to it or something. Should I make it "/console"?

suzannehamilton: Thanks for trying out the URL encoding - it's a shame Jenkins doesn't support it.

I think /input might be the temporary URL that you go to approve or cancel Terraform changes, so it might be correct depending on the context. Which job generated /input URL?

techncl:
The "Approve Terraform plan" stage  - https://github.com/nationalarchives/tdr-terraform-environments/blob/95a8781fec654aff4ba0c99aaf603af692f79091/Jenkinsfile#L57

suzannehamilton:
I just checked, and the /input URL is correct: it's a temporary URL that you can visit when the job is [paused for input](https://www.jenkins.io/doc/pipeline/steps/pipeline-input-step/), in this case so that you can approve or cancel the Terraform changes.

techncl:
Cool. Thanks!

PR #80, opened 2020-07-01
Make environment name available as environment variable

TomJKing:
Each TDR environment has specific Keycloak configuration for some properties which differ across the TDR environments

Require the TDR environment to be available as environment variable to know which environment Keycloak docker image is being built in to apply the specific Keycloak configuration

### DIFF
--- a/modules/keycloak/templates/keycloak.json.tpl
+++ b/modules/keycloak/templates/keycloak.json.tpl
@@ -35,6 +35,10 @@
     ],
     "environment": [
       {
+        "name" : "ENVIRONMENT",
+        "value" : "${app_environment}"
+      },
+      {
         "name" : "FRONTEND_URL",
         "value" : "${frontend_url}"
       },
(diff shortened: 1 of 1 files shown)


modules/keycloak/templates/keycloak.json.tpl
```diff
@@ -34,6 +34,10 @@
       }
     ],
     "environment": [
+      {
+        "name" : "ENVIRONMENT",
```
suzannehamilton, line 38: Will this still be necessary if we can select the right configuration at build time in Jenkins?

I'm always a bit wary of adding an "environment" parameter to an application because the conditional logic within the application tends to get messy, and you end up with an application that contains configuration for other environments (even though it's never used). It's easier to follow the application logic if you can inject the environment-specific configuration instead, like we're doing for the frontend URL just below.

It's just a slight concern. 🙂 If it's necessary then it's fine - it would just be good to consider alternatives.

PR #83, opened 2020-07-09
Tdr 334 add file format resources

MancunianSam:


modules/shared-vpc/root.tf
```diff
@@ -3,7 +3,8 @@ data "aws_availability_zones" "available" {
 
 resource "aws_vpc" "main" {
   cidr_block = "10.0.0.0/16"
-
+  enable_dns_hostnames = true
```
TomJKing, line 6: Has terraform fmt been run?

root.tf
```diff
@@ -309,3 +311,33 @@ module "api_update_lambda" {
...
+  file_system                           = module.file_format_efs.file_system
+  file_format_efs_access_point          = module.file_format_efs.access_point
+  vpc_id                                = module.shared_vpc.vpc_id
+}
+
+module "file_format_efs" {
+  source            = "./tdr-terraform-modules/efs"
+  common_tags       = local.common_tags
+  function          = "file-format-efs"
+  project           = "tdr"
```
TomJKing, line 332: Should this be var.project instead of the hardcoded tdr value?

MancunianSam: Yep, you're right, I've changed it.

modules/shared-vpc/root.tf
MancunianSam: I thought I had, I missed this one.

PR #88, opened 2020-08-06
Add git secrets check to Jenkinsfile

MancunianSam:


Jenkinsfile
```diff
@@ -8,6 +9,16 @@ pipeline {
     choice(name: "STAGE", choices: ["intg", "staging", "prod"], description: "The stage you are deploying Terraform changes to")
   }
   stages {
+    stage("Run git secrets") {
+      agent {
```
suzannehamilton, line 13: Does this mean that the job ends up using two agents on master? I think this is what caused the Jenkins deadlocks.

If you don't specify an agent, does it still work, using the agent that's being used for the whole job?

PR #92, opened 2020-08-28
Move keycloak to postgres

MancunianSam:
The version of mysql on aurora serverless was too old to work with the
latest version of keycloak properly.
We were also getting a lot of connection errors from keycloak which is
unrelated to the upgrade but still isn't great.
So I've moved keycloak to use standard postgres RDS which is the same DB
the consignment api uses so we're at least consistent now.

### DIFF
--- a/modules/keycloak/database.tf
+++ b/modules/keycloak/database.tf
@@ -27,7 +27,7 @@ resource "aws_rds_cluster" "keycloak_database" {
   engine_version                  = "11.6"
   availability_zones              = var.database_availability_zones
   database_name                   = var.app_name
-  master_username                 = "api_admin"
+  master_username                 = "keycloak_admin"
   master_password                 = random_password.password.result
   final_snapshot_identifier       = "keycloak-db-final-snapshot-${random_string.snapshot_prefix.result}-${var.environment}"
   storage_encrypted               = true
@@ -36,10 +36,10 @@ resource "aws_rds_cluster" "keycloak_database" {
   db_subnet_group_name            = aws_db_subnet_group.user_subnet_group.name
   enabled_cloudwatch_logs_exports = ["postgresql"]
   tags = merge(
-  var.common_tags,
-  map(
-  "Name", "content-db-cluster-${var.environment}"
-  )
+    var.common_tags,
+    map(
+      "Name", "keycloak-db-cluster-${var.environment}"
+    )
   )
 
   lifecycle {
--- a/modules/keycloak/database.tf
+++ b/modules/keycloak/database.tf
@@ -22,23 +22,24 @@ resource "random_string" "snapshot_prefix" {
 }
 
 resource "aws_rds_cluster" "keycloak_database" {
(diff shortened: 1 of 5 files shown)


modules/keycloak/variables.tf
```diff
@@ -27,4 +27,6 @@ variable "frontend_url" {}
 variable "ip_whitelist" {
   description = "IP addresses allowed to access"
   default     = ["0.0.0.0/0"]
-}
\ No newline at end of file
+}
+
+variable "kms_key_id" {}
```
suzannehamilton, line 32: Could you just stick a description in here so it's clear it's used as the database encryption key?

modules/keycloak/database.tf
```diff
@@ -22,23 +22,24 @@ resource "random_string" "snapshot_prefix" {
...
-  final_snapshot_identifier = "user-db-final-snapshot-${random_string.snapshot_prefix.result}-${var.environment}"
-  vpc_security_group_ids    = aws_security_group.database.*.id
-  db_subnet_group_name      = aws_db_subnet_group.user_subnet_group.name
-
+  cluster_identifier_prefix       = "keycloak-db-postgres-${var.environment}"
+  engine                          = "aurora-postgresql"
+  engine_version                  = "11.6"
+  availability_zones              = var.database_availability_zones
+  database_name                   = var.app_name
+  master_username                 = "api_admin"
```
suzannehamilton, line 30: Should this be `keycloak_admin`?

modules/keycloak/database.tf
```diff
@@ -22,23 +22,24 @@ resource "random_string" "snapshot_prefix" {
...
+  db_subnet_group_name            = aws_db_subnet_group.user_subnet_group.name
+  enabled_cloudwatch_logs_exports = ["postgresql"]
   tags = merge(
-    var.common_tags,
-    map(
-      "Name", "content-db-cluster-${var.environment}"
-    )
+  var.common_tags,
+  map(
+  "Name", "content-db-cluster-${var.environment}"
```
suzannehamilton, line 41: Should this reference Keycloak or auth server rather than content DB?

Also, something's up with the indentation of the tags.

suzannehamilton (commented):
Just a few small comments - switching to Postgres seems fine to me. 👍 

Can you warn everyone before you deploy to staging, and recreate the admin users in the TDR realm? Check with Sarra too in case she's doing some user research and has created accounts for people.

modules/keycloak/variables.tf
MancunianSam: Yep, I've added one in.

modules/keycloak/database.tf
MancunianSam: It should, can you tell I copied and pasted from the consignment api one :smile:

modules/keycloak/database.tf
MancunianSam: I've changed it to say keycloak and run terraform fmt on the file, it's sorted out the indentation.

suzannehamilton (approved):

PR #96, opened 2020-09-29
Add dependabot config for terraform

MancunianSam:


suzannehamilton:
Out of curiosity, have you found out what sorts of updates this will make? It sounds like [Terraform support is very new](https://dependabot.com/terraform/).

MancunianSam:
I'm not sure to be honest. That's why I wanted to add it, I was curious. :smile:

suzannehamilton (approved):

PR #103, opened 2020-10-12
Update lambda to use EFS

MancunianSam:


root.tf
```diff
@@ -290,7 +295,6 @@ module "checksum_sqs_queue" {
   common_tags              = local.common_tags
   project                  = var.project
   function                 = "checksum"
```
TomJKing, line 297: Is the function variable needed here anymore? Should it be "checksum"?

TomJKing (commented):
Looks good. One minor comment/question

root.tf
MancunianSam: Yeah I think so, it's used to create the queue name
 `sqs_name       = "${var.project}-${var.function}-${local.environment}"`
We still need the checksum queue because that's where the download files lambda is sending its messages.

TomJKing (approved):

PR #105, opened 2020-11-06
Consignment export resources

MancunianSam:


TomJKing (dismissed):

modules/transfer-frontend/variables.tf
```diff
@@ -31,4 +31,6 @@ variable "client_secret_path" {}
 variable "ip_whitelist" {
   description = "IP addresses allowed to access"
   default     = ["0.0.0.0/0"]
-}
\ No newline at end of file
+}
+
+variable "export_api_url" {}
```
TomJKing, line 36: Missing line break

TomJKing (commented):
Looks good. Just one of the usual ;)

modules/transfer-frontend/variables.tf
MancunianSam: I've added this one in

TomJKing (approved):

PR #107, opened 2020-12-14
Add alb network interface private IPs

MancunianSam:
We need to add the trusted IP addresses to the play app so that the
https forwarding from the load balancer works properly.
The IPs that the load balancer uses are the private IPs of the network
interfaces. These are created by AWS when you create the load balancer
and can't be specified in terraform directly so I'm using the network
interface data to get the interfaces for both availability zones and
passing them to the play app via environment variables.

### DIFF
--- a/modules/shared-vpc/root.tf
+++ b/modules/shared-vpc/root.tf
@@ -30,7 +30,8 @@ resource "aws_subnet" "private" {
 }
 
 
-
+# The frontend uses the network interface IPs which come from the range in this subnet.
+# If any changes are made to this subnets IP range, the front end will need to be re-deployed.
 # Create var.az_count public subnets, each in a different AZ
 resource "aws_subnet" "public" {
   count                   = var.az_count
--- a/modules/transfer-frontend/ecs.tf
+++ b/modules/transfer-frontend/ecs.tf
@@ -10,6 +10,8 @@ resource "aws_ecs_cluster" "frontend_ecs" {
   )
 }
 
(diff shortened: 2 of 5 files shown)


suzannehamilton:
Will these IPs change if we reconfigure the VPC subnets?

If so, will that cause issues on the frontend, because it will need to be redeployed for it to pick up the new IP addresses?

MancunianSam:
These are the IPs of the network interfaces so they will change if we make changes to the load balancer and in that case, yes, we'd need to redeploy the front end. But we're unlikely to make changes to the load balancer. The only other way I can think of is to have the trusted IPs as 0.0.0.0/0

suzannehamilton:
What sort of load balancer changes would cause a problem? Anything that redeploys the load balancer? Or only if we make changes to the subnet ranges?

I'm just wondering if we can catch that kind of change before it happens, even if it's in a low-tech way by adding a comment to the relevant place in the Terraform code. Otherwise we risk breaking the service for users until the next time the frontend is deployed.

MancunianSam:
It will be if we redeploy the load balancers or if we make changes to the public subnets that the load balancer sits in. I can't think of any automated way of redeploying as terraform won't redeploy ECS services but I think a comment in the appropriate places makes sense. I've added a comment to the public subnet resource and the front end alb resource.

suzannehamilton (approved):
I'm hoping that if this _does_ cause problems, it will happen in integration and staging first. The risk seems low enough that I'm happy to try it out and see how stable it is.

PR #110, opened 2021-01-22
Add notifications sns topic.

MancunianSam:
The notification lambda will subscribe to this topic in each account so any messages sent to it will be sent to the notifications lambda. The arn for the topic is passed to the step function so it can send messages on success/failure of the export.


suzannehamilton (commented):
This also looks fine, but needs https://github.com/nationalarchives/tdr-terraform-modules/pull/80 to be merged first.

suzannehamilton (approved):

PR #111, opened 2021-01-27
Add an awslogs-datetime-format

MancunianSam:
The problem is that when a stack trace is logged, cloudwatch puts each trace line in its own event which is hard to read. This tells cloudwatch the format of the date to look for and will put any following lines in the same event until it gets to another date.


TomJKing (approved):
Looks good. Just wondering if there is any way to abstract the log config out, so it can be reused. That would also ensure the correct settings are used across different configurations

PR #114, opened 2021-03-04
Create production

MancunianSam:
- Add depends_on for the lambda.
- Remove data blocks from root_data and use the module outputs directly
- Export fields needed by other modules
- Output the mount targets which are used by the lambdas
- Remove the CreateLogGroup permission

### DIFF
--- a/root.tf
+++ b/root.tf
@@ -246,6 +246,8 @@ module "antivirus_lambda" {
   use_efs                                = true
   vpc_id                                 = module.shared_vpc.vpc_id
   private_subnet_ids                     = module.backend_checks_efs.private_subnets
+  mount_target_zero = module.backend_checks_efs.mount_target_zero
+  mount_target_one = module.backend_checks_efs.mount_target_one
 }
 
 module "checksum_lambda" {
@@ -259,6 +261,8 @@ module "checksum_lambda" {
   use_efs                                = true
   backend_checks_efs_root_directory_path = module.backend_checks_efs.root_directory_path
   private_subnet_ids                     = module.backend_checks_efs.private_subnets
+  mount_target_zero = module.backend_checks_efs.mount_target_zero
+  mount_target_one = module.backend_checks_efs.mount_target_one
 }
 
 module "dirty_upload_sns_topic" {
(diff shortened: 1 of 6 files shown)


root.tf
```diff
@@ -315,7 +320,7 @@ module "file_format_sqs_queue" {
   function                 = "file-format"
   dead_letter_queue        = module.backend_check_failure_sqs_queue.sqs_arn
   redrive_maximum_receives = 3
-  visibility_timeout       = 180
+  visibility_timeout       = 900
```
suzannehamilton, line 323: What's the reason for configuring a longer visibility timeout on this queue compared to the checksum one?

root.tf
```diff
@@ -375,6 +383,9 @@ module "backend_checks_efs" {
   access_point_path            = "/backend-checks"
   policy                       = "backend_checks_access_policy"
   mount_target_security_groups = flatten([module.file_format_lambda.file_format_lambda_sg_id, module.download_files_lambda.download_files_lambda_sg_id, module.file_format_build_task.file_format_build_sg_id, module.antivirus_lambda.antivirus_lambda_sg_id, module.checksum_lambda.checksum_lambda_sg_id])
+  nat_gateway_ids              = module.shared_vpc.nat_gateway_ids
+  vpc_cidr_block               = module.shared_vpc.vpc_cidr_block
+  vpc_id                       = module.shared_vpc.vpc_id
```
suzannehamilton, line 388: How does this all work on intg and staging when these weren't configured?

root.tf
MancunianSam: The queue timeout needs to be at least as long as the lambda timeout and the lambda timeout needs to be longer for the file format because of how slow the file format is.

root.tf
MancunianSam: They already existed in integration/staging so when I added the data blocks, it worked because they existed. For a new environment, they don't yet exist so it didn't work. They should never have been added as data blocks, we should always add stuff in as variables so terraform can work out the dependencies.

root.tf
MancunianSam: I'm wary of introducing more dependencies between things as this causes problems when creating and destroying a new environment. 
Also, because there's one file for each lambda, we'd need an output in the lambda module for each lambda timeout like this:
```
output "file_format_timeout" {
  value = aws_lambda_function.file_format_lambda_function.timeout
}

output "checksum_timeout" {
  value = aws_lambda_function.checksum_lambda_function.timeout
}

output "download_files_timeout" {
  value = aws_lambda_function.download_files_lambda_function.timeout
}
... All other lambdas with associated queues
```
That's introducing a lot of output variables. In general it's quite a few changes to avoid something that is easy enough to explain. I could put a comment above this if it helps?

suzannehamilton: OK, in that case a comment sounds good. Just trying to make it easier for people to change things without either accidentally breaking things or worrying about accidentally breaking things. 👍

suzannehamilton (approved):

PR #124, opened 2021-03-17
Add policy to ECS task role

MancunianSam:


modules/consignment-api/ecs.tf
```diff
@@ -1,6 +1,9 @@
 locals {
   app_port = 8080
 }
+
+data "aws_caller_identity" "current" {}
```
TomJKing, line 5: Is it worth putting this in a `data.tf`, to keep things consistent?

modules/consignment-api/ecs.tf
```diff
@@ -78,6 +80,16 @@ resource "aws_iam_role" "consignment_api_ecs_execution" {
   )
 }
 
+resource "aws_iam_role_policy_attachment" "assume_iam_auth_attach" {
```
TomJKing, line 83: Don't think the `_attach` is needed on the end of the name, for consistency

modules/consignment-api/templates/ecs_assume_role_policy.json.tpl
```diff
@@ -0,0 +1,13 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Action": "sts:AssumeRole",
+      "Principal": {
+        "Service": "ecs-tasks.amazonaws.com"
+      },
+      "Effect": "Allow",
+      "Sid": ""
```
TomJKing, line 10: Is the `Sid` field needed, as there is no value for it?

TomJKing (commented):
Looks good. Couple of minor comments

modules/consignment-api/ecs.tf
MancunianSam: Yeah, should do, I was being lazy.

modules/consignment-api/ecs.tf
MancunianSam: I've taken it off.

modules/consignment-api/templates/ecs_assume_role_policy.json.tpl
MancunianSam: No you're right, bad copy and paste job.

TomJKing (approved):

PR #125, opened 2021-03-17
Add permissions to allow migrations lambda to use iam db auth

MancunianSam:


modules/database-migrations/templates/migration_lambda.json.tpl
```diff
@@ -0,0 +1,38 @@
...
+      "Effect": "Allow",
+      "Action": [
+        "ec2:DeleteNetworkInterface",
+        "ec2:DescribeInstances",
+        "ec2:CreateNetworkInterface",
+        "ec2:AttachNetworkInterface",
+        "ec2:DescribeNetworkInterfaces",
+        "autoscaling:CompleteLifecycleAction"
+      ],
+      "Resource": ["*"]
```
suzannehamilton, line 35: I know this line has just been moved from elsewhere, but do we need to revisit this and make it more specific so the migration lambda can't attach any network interface?

suzannehamilton (approved):

PR #126, opened 2021-03-25
Add GovUk Notify params

TomJKing:
Depends on PR: https://github.com/nationalarchives/tdr-terraform-backend/pull/105

Parameters to support GovUk Notify as email service for Keycloak

### DIFF
--- a/modules/keycloak/ssm.tf
+++ b/modules/keycloak/ssm.tf
@@ -81,3 +81,29 @@ resource "aws_ssm_parameter" "keycloak_user_admin_client_secret" {
     ignore_changes = [value]
   }
 }
+
+//No programmatical method to retrieve value from GoUk Notify service
+//Use Terraform to add the parameter and update value manually
+//As added by Terraform won't be removed when Terraform apply run
+resource "aws_ssm_parameter" "keycloak_govuk_notify_api_key" {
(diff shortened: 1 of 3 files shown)


modules/keycloak/ssm.tf
```diff
@@ -81,3 +81,29 @@ resource "aws_ssm_parameter" "keycloak_user_admin_client_secret" {
     ignore_changes = [value]
   }
 }
+
+//No programmatical method to retrieve value from GoUk Notify service
+//Use Terraform to add the parameter and update value manually
+//As added by Terraform won't be removed when Terraform apply run
+resource "aws_ssm_parameter" "keycloak_govuk_notify_api_key" {
```
suzannehamilton, line 88: Just a note to myself to come back to this once we've discussed https://github.com/nationalarchives/tdr-terraform-backend/pull/105/files#r603079753 - I'm still trying to wrap my head around how this parameter gets set in the different environments.

suzannehamilton (approved):
In case this isn't already on your list of PRs, it would be good to add the Notify parameters to the list of manual setup steps here:  https://github.com/nationalarchives/tdr-dev-documentation/blob/master/manual/create-new-environment.md

PR #127, opened 2021-03-29
Change KMS key policy and pass kms arn to lambdas

MancunianSam:
The lambdas need to have their environment variables encrypted in transit using a KMS key. The encrypted environment variables are set using terraform and the lambda has to decrypt them. There are two steps to allow the lambda to decrypt the environment variables using KMS. The lambda execution policy needs permission to decrypt using the key and the key policy itself needs to allow the lambda to decrypt the files. Most of the changes are in https://github.com/nationalarchives/tdr-terraform-modules/pull/92 which should make this a bit clearer.

### DIFF
--- a/root.tf
+++ b/root.tf
@@ -207,6 +207,7 @@ module "frontend_alb" {
 module "encryption_key" {
   source      = "./tdr-terraform-modules/kms"
   project     = var.project
+  key_policy  = "lambda"
   function    = "encryption"
   environment = local.environment
   common_tags = local.common_tags
@@ -246,6 +247,7 @@ module "antivirus_lambda" {
   private_subnet_ids                     = module.backend_checks_efs.private_subnets
   mount_target_zero                      = module.backend_checks_efs.mount_target_zero
   mount_target_one                       = module.backend_checks_efs.mount_target_one
+  kms_key_arn                            = module.encryption_key.kms_key_arn
 }
 
 module "checksum_lambda" {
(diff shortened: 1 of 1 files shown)


suzannehamilton:
Could you explain in the commit message why this is necessary? Changing an encryption policy seems like a pretty major change! Is this changing it in a way that affects all our use of KMS?

MancunianSam:
I've updated the PR description. Most of the changes are in the tdr-terraform-modules project, it's a bit more obvious in there what's happening.

MancunianSam:
Thinking about this, the reason I'm updating the key policy is because I'm reusing an existing KMS key. This is the key that encrypts the two databases. Now I think about it, we should probably have a separate key to do the lambda environment variables. What do you think?

MancunianSam:
I found a [best practices document](https://d0.awsstatic.com/whitepapers/aws-kms-best-practices.pdf) from AWS  and this is what they say:

> Since AWS makes creating and managing keys easy through the use of AWS
> KMS, we recommend that you have a plan for how to use the service to best
> control the blast radius around individual keys. Previously, you may have used
> the same key across different geographic regions, environments, or even
> applications. With AWS KMS, you should define data classification levels and
> have at least one CMK per level. For example, you could define a CMK for data
> classified as “Confidential,” and so on. This ensures that authorized users only
> have permissions for the key material that they require to complete their job.

So I guess the question is do we want to restrict access to some encrypted things, databases, lambda environment variables, sqs queues etc or is this unnecessarily complicated?

suzannehamilton:
Can you post this on Slack? Let's see what David thinks.

MancunianSam:
David is happy using a single key so this PR is good to go.

suzannehamilton (approved):

PR #131, opened 2021-05-11
Fetch WAF allow list from tdr-configurations instead of SSM param

suzannehamilton:
This changes where the list of IPs which are allowed to access Keycloak admin URLs is read from.

Before, we read the list from a parameter in the AWS SSM Parameter Store. Now, it reads the list from the tdr-configurations repo.

The tdr-configurations repo was already being used for the frontend load balancer allow list, so this change makes the WAF consistent with the frontend.

It also reduces the number of deployment steps. When the list changes, we can now deploy the changes by running this Terraform project, without having to deploy the change to the parameter store first.

Also rename "whitelist" to "allow list" to match the GOV.UK styleguide: https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#allow-list

### DIFF
--- a/modules/consignment-api/security.tf
+++ b/modules/consignment-api/security.tf
@@ -54,7 +54,7 @@ resource "aws_security_group" "lb" {
     protocol    = "tcp"
     from_port   = 443
     to_port     = 443
-    cidr_blocks = var.ip_whitelist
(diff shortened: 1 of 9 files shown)


MancunianSam (approved):
Looks good, thanks

PR #144, opened 2021-06-17
Add rules for tdr public subnets.

MancunianSam:
The main difference between security groups and NACLs is that security groups are stateful. The security group for the load balancer allows access to everything outbound. If the security group allows a connection outbound, it will allow the return connection as well.
There is more detail on the difference here https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Security.html#VPC_Security_Comparison

NACLs are stateless which means that rules which allow inbound traffic need to be stated explicitly. So if we were trying to restrict by IP address, we would need to add the IP address for the API, for keycloak ,the consignment export and for s3. At the moment, we're talking to S3 over the public internet and there's no definitive list of S3 IPs as I remember so we can't restrict the cidr range here or the s3 upload won't work.

In the future, if we use s3 vpc endpoints, we may be able to revisit this but it's not going to provide much in the way of extra security because we already have the security group configured properly.

The rules for the ephemeral ports are explained here https://docs.aws.amazon.com/vpc/latest/userguide/vpc-network-acls.html#nacl-ephemeral-ports
We need specific ports like 9000 for the Play app but we need the others for the load balancer health checks.

There are some more examples here of rule setup for VPCs like ours https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Scenario2.html#nacl-rules-scenario-2

The rules here are more general than I would like because we have so many resources inside the VPC going out to the public internet for things like SSM, S3, ECR etc. and opening up port 443 inbound doesn't seem to work so AWS must be using non-standard ports for things like this.

Ideally, for the private subnets, we should create VPC endpoints for all of the services which we need and then we can restrict access to requests from inside the VPC but that's quite a large bit of work which we can do later.


TomJKing (approved):
Looks OK.

PR #147, opened 2021-07-06
Increase the message visibility timeout for file checks

suzannehamilton:
Set the SQS message visibility timeout to be three times the lambda timeout for each of the file check lambdas, including the file download and API update lambda.

This should prevent messages from being sent to the failure queue incorrectly. See ADR 20 for how this change works: nationalarchives/tdr-dev-documentation#122

This will make retries slower, because the message visibility timeout has to expire before a failed file check is retried. The plan is to add extra error handling to the file checks which sets the message visibility timeout to zero to cause the check to be retried immediately.

Marking as a draft PR until https://github.com/nationalarchives/tdr-terraform-modules/pull/124 is merged.

### DIFF
--- a/root.tf
+++ b/root.tf
@@ -242,7 +242,7 @@ module "antivirus_lambda" {
   common_tags                            = local.common_tags
   file_system_id                         = module.backend_checks_efs.file_system_id
   lambda_yara_av                         = true
-  timeout_seconds                        = 180
+  timeout_seconds                        = local.file_check_lambda_timeouts_in_seconds["antivirus"]
   project                                = var.project
   use_efs                                = true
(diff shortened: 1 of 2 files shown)


TomJKing (commented):
Looks good to me. Might be worth getting @MancunianSam opinion as well.

MancunianSam:
Looks good. It's nice having the timeouts in one place.

suzannehamilton:
The modules PR has been merged, so I've marked this as ready for review. Could someone approve it if they're happy with it?

@TomJKing What's the best way to get this into your v13 branch?

TomJKing (approved):

PR #151, opened 2021-07-13
Upgrade aws provider 3

TomJKing:


Jenkinsfile
```diff
@@ -6,5 +6,6 @@ terraformDeployJob(
   taskRoleName: "TDRTerraformAssumeRole${params.STAGE.capitalize()}",
   deployment: "Environment",
   terraformDirectoryPath: ".",
-  testDelaySeconds: 300
+  testDelaySeconds: 300,
+  modulesBranch: "update-aws-provider-version-support"
```
MancunianSam, line 10: Is this a temporary thing while you're moving stuff over or does it need to go in with this PR?

TomJKing: It needs to be there until all the other repos are updated, to provide backwards compatibility. Once everything is upgraded I'll merge the changes from the branch in the tdr-terraform-modules repo, and then remove this in another PR. Only way to do it without breaking everything.

MancunianSam (approved):

PR #156, opened 2021-08-20
Add bastion sg to inbound rules for both efs volumes.

MancunianSam:
This will allow us to use the bastion to connect to the EFS volumes.
There have been a few times recently where we've needed to so this
removes the need to manually fiddle with the security groups.

### DIFF
--- a/modules/consignment-api/outputs.tf
+++ b/modules/consignment-api/outputs.tf
@@ -30,3 +30,7 @@ output "database_cluster_id" {
 output "database_security_group" {
   value = aws_security_group.database.id
 }
+
+output "bastion_security_group_id" {
+  value = aws_security_group.bastion_security_group.id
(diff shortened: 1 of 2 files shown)


suzannehamilton:
I think that having a bastion to access EFS is a good idea, thanks.

There are a few security questions I think it would be good to think about and discuss with David: https://national-archives.atlassian.net/browse/TDR-1319 It's things like whether it should be a separate bastion (so we're not spinning up an instance that has access to everything every time) and whether it's possible to make a read-only connection using a separate mount point.

MancunianSam:
We're going to separate these permissions out into two roles which will be attached to the bastion separately depending on what we're using it for. Or we'll have two separate  bastions. Either way, the role changes will be different so this can be closed.

PR #162, opened 2021-09-22
Add signed cookies API and lambda

MancunianSam:
I've added in the signed cookies API gateway resource and the lambda that will respond to requests from API gateway.


TomJKing (approved):
Looks good.

PR #164, opened 2021-09-28
Remove Cognito and original bucket notification

MancunianSam:
Remove Cognito
    
We're not using it any more, we're using signed cookies with Cloudfront instead so it can be removed. 

I've also Remove the SNS notification from the original dirty bucket. Once we've switched over to use the new cloudfront upload url, we won't  need the notification from the old bucket. It's probably worth keeping the bucket for now as there's things in it on prod but eventually we should be able to get rid.


TomJKing (approved):
Satisfying ;) Probably worth adding a JIRA task to the backlog about removing the old bucket at some point.

MancunianSam:
I've created one here https://national-archives.atlassian.net/browse/TDR-1394

PR #167, opened 2021-11-03
Create Keycloak in TDR VPC

MancunianSam:
This uses new tdr-terraform-modules to create the keycloak resource inside the root.tf file instead of having it in its own separate module.


TomJKing (commented):
Looks good. One comment. Added most of my questions to the tdr-terraform-modules PR: https://github.com/nationalarchives/tdr-terraform-modules/pull/149

root.tf
```diff
@@ -776,3 +776,183 @@ module "bastion_role" {
   name               = "BastionEC2Role${title(local.environment)}"
   policy_attachments = {}
 }
+
```
TomJKing, line 779: This is getting quite large now. Is it possible to split this file up?

MancunianSam: Yeah, makes sense, I've moved the new Keycloak resources into a new file

TomJKing (approved):

PR #168, opened 2021-11-08
Move the route53 entry to the keycloak tdr alb

MancunianSam:
This will need some manual intervention in that before it's run, the
state will have to be moved from one resource to another.

terraform state mv module.keycloak.aws_route53_record.keycloak_dns 'module.keycloak_route53.aws_route53_record.dns[0]'

This will stop terraform destroying the record and us having to wait for
the DNS to propagate again.


root_keycloak.tf
```diff
@@ -177,3 +177,16 @@ module "create_keycloak_db_users_lambda_new" {
...
+  common_tags           = local.common_tags
+  environment_full_name = local.environment_full_name
+  project               = "tdr"
+  a_record_name         = "auth"
+  alb_dns_name          = module.keycloak_tdr_alb.alb_dns_name
+  alb_zone_id           = module.keycloak_tdr_alb.alb_zone_id
+  create_hosted_zone    = false
+  hosted_zone_id        = data.aws_route53_zone.tdr_dns_zone.id
+
+}
```
TomJKing, line 192: Missing line break

TomJKing (commented):
One very minor comment. Assume the manual intervention on the Terraform state will be a one off operation.

MancunianSam:
It will be a one off state move yeah, after it's done the new resource will point to the existing record permanently.

TomJKing (approved):

PR #170, opened 2021-11-09
Delete old keycloak module

MancunianSam:
- Move consignment export task.
- Change names and descriptions to match existing resources.
- Delete old keycloak module.


TomJKing (commented):
Looks good. Have the staging and prod environments been moved yet? Wondering if this is safe to do until they've been migrated aswell?

TomJKing (approved):

PR #182, opened 2022-02-14
Bulk create keycloak users

MancunianSam:
- Add modules for lambda, api gateway and S3
- Add API Gateway swagger json


templates/api_gateway/create_keycloak_users.json.tpl
```diff
@@ -0,0 +1,70 @@
+{
+  "openapi" : "3.0.1",
+  "info" : {
+    "title" : "TestKeycloakCreateUsers",
```
TomJKing, line 4: Should this have "Test" in the title?

TomJKing (commented):
Looks good. One minor comment

templates/api_gateway/create_keycloak_users.json.tpl
MancunianSam: No, I've changed it.

TomJKing (approved):

PR #186, opened 2022-03-01
Change e2e tests environment

MancunianSam:
* Remove the team slug to remove environment restrictions so e2e test
  runs don't need to be approved.
* Don't create an environment for prod as we don't run the tests on
  prod.


TomJKing (approved):
Looks good. Thanks

PR #188, opened 2022-03-10
Add frontend environment and update consignment api

MancunianSam:
I've changed the github actions so they don't need as many secrets and
so they don't have to run in the repository environment for the tests
which causes all sorts of noise on the pull request. I've updated the
API one and added in one for the front end so I can move that over to
github actions.

### DIFF
--- a/root_github.tf
+++ b/root_github.tf
@@ -12,12 +12,7 @@ module "github_consignment_api_environment" {
   repository_name = "nationalarchives/tdr-consignment-api"
   team_slug       = "transfer-digital-records-admins"
   secrets = {
-    TITLE_STAGE            = title(local.environment)
-    ACCOUNT_NUMBER         = data.aws_caller_identity.current.account_id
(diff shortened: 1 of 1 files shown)


root_github.tf
```diff
@@ -38,6 +33,16 @@ module "github_e2e_tests_environment" {
   }
 }
 
+module "github_transfer_frontend_environment" {
+  source          = "./tdr-terraform-modules/github_environments"
+  environment     = local.environment
+  repository_name = "nationalarchives/tdr-file-format"
```
TomJKing, line 39: Is that the correct repo name?

TomJKing (commented):
Looks good. Just one minor comment.

root_github.tf
MancunianSam: Nope :grin: I was testing it on that repo and forgot to change it.

TomJKing (approved):

PR #190, opened 2022-03-15
Change e2e tests secrets.

MancunianSam:
We want the secrets to be on the repository rather than in a GitHub
environment. Environments are meant to be for deploying code so any PR
where the tests are running currently shows a lot of spam from GitHub
about a deployment starting.

We still need environment specific secrets but we're doing this by
prefixing them with the environment and adding them at the repository
level.


TomJKing (approved):
Looks good.

PR #193, opened 2022-03-17
Setup tdr-checksum for github actions

MancunianSam:
I've added the lambda to the policy and add the ACCOUNT_NUMBER to the environment secrets.


TomJKing (approved):
Looks good.

PR #199, opened 2022-04-06
Add auth server secrets

MancunianSam:
- Replace test branch with main
- Add auth server secrets.


TomJKing (approved):
Looks good. Thanks

PR #200, opened 2022-04-08
Add resources to update Keycloak config.

MancunianSam:
This is the ECS task and roles and policies along with the role for
GitHub to assume to run the ECS task.


TomJKing (approved):
Looks good. Thanks

PR #205, opened 2022-04-26
Update release versions.

MancunianSam:
We're not updating the release versions like the Jenkins job used to do
so we can't see when the terraform run is out of date.


TomJKing (approved):
Looks good. Thanks

PR #209, opened 2022-04-28
Add policies to create bastion.

MancunianSam:
There's no need to use the old ones as we can assume a role directly
into the target account rather than assuming a role from management.


TomJKing (approved):
Looks good. Thanks

PR #210, opened 2022-04-28
Add lambdas to GitHub policy

MancunianSam:
This is the policy that allows us to deploy the lambdas. I was adding
them one at a time but I keep forgetting so I've added in all the
missing ones in one go.


TomJKing (approved):
Looks good. 

Wondering whether its worth just wild carding the Resources? Given that Github actions is going to need to deploy any TDR lambdas? `"arn:aws:lambda:${region}:${account_id}:function:tdr-*`

PR #218, opened 2022-05-09
Add file format roles and secrets.

MancunianSam:
The role needed by the file format run ECS task job is very similar to
the one for running the Keycloak update job so I've refactored the
keycloak one to work with both.


TomJKing (approved):
Looks good. Thanks

PR #223, opened 2022-05-19
Add AUTH_URL environment variable.

MancunianSam:
This is mostly for the upgrade to Keycloak so I can change the auth
server url by changing terraform but it's useful to have everything as
an environment variable in case it changes again.


TomJKing (approved):
Looks good. Thanks

PR #224, opened 2022-05-19
Upgrade Keycloak

MancunianSam:
- Change the environment variable names
- Add the keycloak host
- Update the auth url everything else uses


TomJKing (approved):
Looks good. Thanks

PR #241, opened 2022-07-18
Remove port 80 from frontend ALB security group.

MancunianSam:
We don't need port 80 open any more as all traffic to the front end is
https so this can be removed and it might help with the problems we have
with the e2e tests.


TomJKing (approved):
Looks good. Thanks

PR #242, opened 2022-07-19
Pass in parameter store variable.

MancunianSam:
This passes in the path to be used as an environment variable as part of
this ticket https://national-archives.atlassian.net/browse/TDR-2179


TomJKing (approved):
Looks good. Thanks

PR #243, opened 2022-08-02
Add rotate secrets lambda

MancunianSam:
- Add rotate secrets lambda
- Update modules commit hash


TomJKing (approved):
Looks good. Thanks.

Regarding the frequency of the rotation. Has the Keycloak client secret expiration policy got sufficient time for a fix if the rotation failed, ie the old secret expiration is greater than 7 days?

MancunianSam:
I haven't done the Keycloak config yet but I'll set it to 10 days ish which should give us a few days to sort things.

PR #244, opened 2022-08-04
Add an environment variable for the rotate secrets client.

MancunianSam:
This is needed when we start the service, it will read the value of the
secret from this environment variable.


TomJKing (approved):
Looks good. Thanks

PR #248, opened 2022-08-11
Move API and Keycloak to another subnet.

MancunianSam:
I wanted to move everything to a new larger subnet but it turns out
that is a lot more difficult than it sounds so for now, I've moved the
API and Keycloak ECS tasks to the backend-checks subnet.

This **should** be a temporary change. There is a ticket
https://national-archives.atlassian.net/browse/TDR-2253
to sort this properly.


TomJKing (approved):
Looks good. Thanks.

PR #249, opened 2022-08-12
TDR 2222 single db instance

MancunianSam:
Add a new single instance database.

This does the following.

Creates a new single instance DB for keycloak and the API.

Updates the create-users lambdas with the new credentials. We don't use these lambdas normally so this is fine.

Update the task roles to allow access to the new databases.

Adds the existing keycloak master credentials to SSM as for some reason they're not there. These will be deleted after this is all deployed.

Once this is merged and deployed, I'll put in another PR to switch over the API and Keycloak to use these DBs.

This is a draft because it needs https://github.com/nationalarchives/tdr-terraform-modules/pull/204 to be merged so I can update the modules hash


TomJKing (dismissed):
Looks good. Thanks

TomJKing (approved):
Looks good. Thanks

PR #251, opened 2022-08-12
Delete old databases

MancunianSam:
Delete the old databases.
    
When this is run, Keycloak and the API will have been moved to the new single instance databases.

I've moved the Keycloak policies into here to try and remove TDR specific stuff from terraform-modules.


templates/iam_policy/keycloak_ecs_task_role_policy.json.tpl
```diff
@@ -0,0 +1,19 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": [
+        "sns:Publish",
```
TomJKing, line 7: Could you order the actions alphabetcially.

TomJKing (commented):
Looks good. One very minor comment.

TomJKing (approved):
Looks good. Thanks

PR #253, opened 2022-08-25
Add upload bucket to file format lambda

MancunianSam:
This will be used by the new file format lambda to get the files directly from S3. This will be released once the droid changes are in but there's no harm in having the variable sat here unused for now.


TomJKing (dismissed):
Looks good. Thanks

TomJKing (approved):
Looks good. Thanks

PR #255, opened 2022-08-26
Use account specific parameter names.

MancunianSam:
The run report action needs to be run by users outside TDR. At the
moment, it's run in an environment which needs approvals.

I want to run it outside the environment but for this we need three
secrets, INTG_ACCOUNT_NUMBER, STAGING_ACCOUNT_NUMBER,
PROD_ACCOUNT_NUMBER which this sets.


TomJKing (approved):
Looks good. Thanks

PR #261, opened 2022-09-06
Add policy for e2e tests.

MancunianSam:
The e2e tests need to get the secrets from the parameter store as when
they're rotated, the hard coded ones in GitHub secrets don't update.

I've added a policy and role to allow it to get the two secrets it needs
and removed the secrets from the environment.


TomJKing (approved):
Looks good. Thanks

PR #280, opened 2022-10-26
Only subscribe to TRE SNS topic in intg

TomJKing:
TRE SNS topics for other environments not currently configured and therefore cannot subcribe to non-existent SNS topics

When the TRE AWS infrastructure for other environments is in place, this should be updated to allow the subscriptions


thanhz (approved):
Looks good

PR #286, opened 2022-11-15
Add waf logging and log group

MancunianSam:
This creates a log group for the waf logs and adds it to the waf module.

The name has to be `aws-waf...` because of AWS rules.


TomJKing (approved):
Looks good. Thanks

PR #294, opened 2023-01-11
TDR-2777 Add feature flag for view history page

thanhz:


root_locals.tf
```diff
@@ -86,4 +86,5 @@ locals {
   //Feature access blocks
   block_feature_closure_metadata     = true
   block_feature_descriptive_metadata = true
+  block_feature_view_history         = true
```
TomJKing, line 89: I think it is safe to release the feature on intg and staging as it won't break the e2e tests, so you can set this as follows: `block_feature_view_transfers = local.environment == "prod" ? true : false`

TomJKing (commented):
Looks good. Just one minor comment about releasing the feature on intg and staging, which I think it is safe to do so.

Also the feature has been renamed to "view transfers" in the  so the code should be updated to reflect this, ie `...view_transfers`

TomJKing (approved):
Looks good. Thanks

PR #296, opened 2023-01-12
Split large log files before sending to Cloudwatch

MancunianSam:
If the plan log is over 262144 bytes, you can't send it in a single log
event. This checks the size and splits it in half if it's too big.

This should keep us going for a while, hopefully until we can split out
tdr-terraform-environments.


TomJKing (approved):
Looks good. Thanks

PR #297, opened 2023-01-12
Move API gateway definitions and policies.

MancunianSam:
These have been moved from tdr-terraform-modules to make the module less
TDR specific and the call to the module has been updated to use the new
modules. This will have to be deployed in tandem with https://github.com/nationalarchives/tdr-terraform-modules/pull/236

If I move the resources using `terraform state mv` this shouldn't need
downtime.


TomJKing (approved):
Looks OK. Thanks

PR #301, opened 2023-01-17
Add an environment variable for the backend checks url

MancunianSam:
We need to call the new backend checks URL to trigger the step function.
This adds it to the config. This can be deployed any time because the
frontend will ignore it if it's not being used.


TomJKing (approved):
Looks good. Thanks

PR #303, opened 2023-01-25
Remove EFS, old lambdas and queues

MancunianSam:
This will remove the old backend checks lambdas, the backend checks EFS volume and the queues and topics we had for the old backend checks.

I had to move the subnet to another module because of the daft way the EFS module was set up. It will be very difficult to delete and recreate a subnet while things are using it so to save a lot of time, this will need a state change before deploying.

```
terraform state mv 'module.backend_checks_efs.aws_subnet.efs_private[0]' 'module.shared_vpc.aws_subnet.private_backend_checks[0]'
terraform state mv 'module.backend_checks_efs.aws_subnet.efs_private[1]' 'module.shared_vpc.aws_subnet.private_backend_checks[1]'
```


TomJKing (approved):
Looks good. Thanks. If its practical when you do the resource `mv` in the terraform state, maybe you can demo this to Vimlesh? I think it would be a good thing to show him.

PR #305, opened 2023-02-02
Set reserved concurrency.

MancunianSam:
This was set to 1 which only allows one to run at a time. This switches
it off.


TomJKing (approved):
Looks good. Thanks

PR #306, opened 2023-02-03
Remove security audit IAM user group

TomJKing:
IAM user group is no longer needed as pen testing complete

Leaving an unsed IAM user group is considered a moderate secruity risk in itself

### DIFF
--- a/root_locals.tf
+++ b/root_locals.tf
@@ -84,7 +84,7 @@ locals {
   transform_engine_v2_sqs_topic_subscriptions = local.environment == "intg" ? [nonsensitive(data.aws_ssm_parameter.transform_engine_v2_tre_out_topic_arn.value)] : []
 
   //Set to true to create security audit IAM user group
-  security_audit = local.environment == false
+  security_audit = false
 
   //Feature access blocks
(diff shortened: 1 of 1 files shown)


root_locals.tf
```diff
@@ -83,8 +83,8 @@ locals {
   //This should be removed once rest of TRE environments are configured
   transform_engine_v2_sqs_topic_subscriptions = local.environment == "intg" ? [nonsensitive(data.aws_ssm_parameter.transform_engine_v2_tre_out_topic_arn.value)] : []
 
-  //Don't run testing on prod
-  security_audit = local.environment == "prod" ? false : true
+  //Set to true to create security audit IAM user group
+  security_audit = local.environment == false
```
vimleshtna, line 87: Correct me if I'm wrong, `local.environment` will be always "intg", "staging" or "prod" not the `true` or `false`.

vimleshtna (commented):
Looks OK to me, just confirming one thing about the `local.environment`

root_locals.tf
TomJKing: Good point!

vimleshtna (approved):
Looks good, thanks.

PR #309, opened 2023-02-06
Add new E2E test policy.

MancunianSam:
For some reason we were assuming a role in GitHub actions in the
management account and then assuming a role into the integration or
staging accounts within the tests. This was necessary when we were
running it in Jenkins but it isn't any longer.

This is a new role which the e2e tests will assume in the actions
runner.


TomJKing (approved):
Looks good. Thanks.

PR #311, opened 2023-02-06
Remove the http listener for the frontend load balancer.

MancunianSam:
We don't need this and the pen test report flagged it.


TomJKing (approved):
Looks good. Thanks

PR #312, opened 2023-02-07
Fix the assume role policy paths.

MancunianSam:
The module which contained the policy json has been deleted. This
updates it to use the json in this repo.


TomJKing (approved):
Looks good. Thanks

PR #313, opened 2023-02-08
Remove the old db security group.

MancunianSam:
This was flagged by the pen testers as unused. I missed this on the
first pass of deleting things but it can go now.


TomJKing (approved):
Looks good. Thanks

PR #315, opened 2023-02-14
Add lambda failure information to backend checks.

MancunianSam:
This has been changed so we can tell when an individual file check has
failed for a particular file.

If there is an error with a file check, we send a message to the
notifications lambda and then add the name of the file check to the
error.
This is then aggreagated at the end of the map function so we should see
"failures": ["Antivirus"] for a particular file. We can use these to set
the file and consignment statuses when the backend checks fail.

This needs to be done first because the existing lambdas will ignore
this new field but it needs to be there before I can change the lambda
code.

I've also removed the general catch block around the map loop because
it's not necessary now we have individual catches on each of the checks
lambdas.

We may need to do some additional error handling on the step function as
a whole but that's a change for another time.

### DIFF
--- a/templates/step_function/backend_checks_definition.json.tpl
+++ b/templates/step_function/backend_checks_definition.json.tpl
@@ -66,19 +66,38 @@
                       "FunctionName": "${yara_av_v2_lambda_arn}",
                       "Payload.$": "$"
                     },
-                    "Retry": [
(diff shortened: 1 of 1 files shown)


TomJKing (approved):
Looks good. Thanks.

PR #320, opened 2023-03-16
Missed a set-output command

MancunianSam:
I didn't notice this, it was picked up by someone else.


TomJKing (approved):
Looks good. Thanks.

PR #323, opened 2023-04-04
TDR-3129 Add x-ray collector container

thanhz:
Update the ECS task to run the collector in the same service

### DIFF
--- a/modules/transfer-frontend/templates/frontend.json.tpl
+++ b/modules/transfer-frontend/templates/frontend.json.tpl
@@ -1,5 +1,28 @@
 [
   {
+    "name": "aws-otel-collector",
+    "image": "229554778675.dkr.ecr.eu-west-2.amazonaws.com/collector",
+    "cpu": 256,
+    "memory": 512,
+    "portMappings": [],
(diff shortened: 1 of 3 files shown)


modules/transfer-frontend/templates/frontend.json.tpl
```diff
@@ -1,4 +1,27 @@
...
+      "--config=/etc/ecs/custom-config.yml"
+    ],
+    "environment": [],
+    "mountPoints": [],
+    "volumesFrom": [],
+    "logConfiguration": {
+      "logDriver": "awslogs",
+      "options": {
+        "awslogs-create-group": "True",
+        "awslogs-group": "/ecs/ecs-aws-otel-sidecar-collector-${app_environment}",
```
thanhz, line 19: should the awslogs-group value be changed to something else?

TomJKing (commented):
Looks good. Thanks. We should hold off deploying these changes until we have an image in the ECR repo, otherwise it'll cause an error.

TomJKing (approved):
Looks good. Thanks

PR #329, opened 2023-04-27
Add correct permissions to task role

TomJKing:


thanhz (approved):
Thanks, looks good to me

PR #332, opened 2023-05-03
TDR-3129 - Add otel_service_name environment variable

thanhz:
Currently the frontend ecs node appears as `unknown_service:java` on the xray ui. This adds a new environment variable `otel_service_name` to the frontend ecs task. So that it appears as shown here:

image


TomJKing (approved):
Looks good. Thanks

PR #338, opened 2023-06-16
Update secrets for custodian and config repos.

MancunianSam:
Now that these repos are being used by other teams, the environments
need to be prefixed with the team name


TomJKing (approved):
Looks good. Thanks

PR #342, opened 2023-06-28
Add 'da-terraform-configuration' as sub-module

TomJKing:
'da-terraform-configurations' hold cross account configuration which is needed for different services to interact with each other

This will be needed for the implementation of the new TRE architecture


thanhz (approved):
Looks good :+1:

PR #345, opened 2023-06-30
Remove github environment configurations as we have migrated these to new tdr-terraform-github repo

vimleshtna:


TomJKing (commented):
Looks good. Thanks. I'm going to hold off approving for now as want to as want to do the migration manually first, and will need to roll back if it goes wrong.

TomJKing (approved):

PR #352, opened 2023-07-31
TRD-2628 Http4s feature block

thanhz:


root_locals.tf
```diff
@@ -105,4 +105,6 @@ locals {
   tmp_directory         = "/tmp"
   consignment_user_name = "consignment_api_user"
 
+  //feature access blocks
+  block_http4s = local.environment == "prod" ? true : false
```
TomJKing, line 109: To be on the safe side, should probably block on `staging` as well. Strictly speaking `staging` and `prod` should be kept insync as far as possible.

TomJKing (commented):
Looks good. Just one very minor thing. Thanks

root_locals.tf
```diff
@@ -105,4 +105,6 @@ locals {
   tmp_directory         = "/tmp"
   consignment_user_name = "consignment_api_user"
 
+  //feature access blocks
+  block_http4s = local.environment == "prod" || local.environment == "staging" ? true : false
```
TomJKing, line 109: Could rewrite this to: 

`block_http4s = local.environment == "intg" ? false : true`

But how it is will be fine, as this is temporary code anyway.

TomJKing (approved):
Looks good. Thanks.

PR #358, opened 2023-09-05
Remove athena s3 bucket and queries

TomJKing:
These are defined in the tdr-aws-accounts: https://github.com/nationalarchives/tdr-aws-accounts/blob/24930201d653e0f9fb27accf84a836efdf0fbd38/root_main.tf#L119

Remove to prevent duplication


vimleshtna (approved):
Looks good, thanks

PR #360, opened 2023-09-14
Local to set the correct tre environment name

TomJKing:
TRE naming of environments does not following TDR's naming convention

Add local to convert the TDR environment name to the correct TRE environment name


vimleshtna (approved):
Looks good, thanks

PR #363, opened 2023-09-20
Cipher text encrytion requires string value

TomJKing:
Until actual event bus arn available use a placeholder value for Terraform to apply successfully


vimleshtna (approved):
Looks good, thanks

PR #369, opened 2023-10-02
Add KMS encryption to s3 export buckets

vimleshtna:


root.tf
```diff
@@ -490,13 +490,15 @@ module "export_bucket" {
   project     = var.project
   function    = "consignment-export"
   common_tags = local.common_tags
+  kms_key_id  = module.s3_external_kms_key.kms_key_arn
```
TomJKing, line 493: Only apply the encryption key to the judgment bucket, not the standard bucket. For the standard bucket need to ensure that the Digital Archivists can still access it after encryption, otherwise they can't process the standard transfers

root.tf
```diff
@@ -490,13 +490,15 @@ module "export_bucket" {
...
 }
 
 module "export_bucket_judgment" {
   source      = "./tdr-terraform-modules/s3"
   project     = var.project
   function    = "consignment-export-judgment"
   common_tags = local.common_tags
+  kms_key_id  = module.s3_external_kms_key.kms_key_arn
```
TomJKing, line 501: As an extra control over the release of this should only add the encryption in intg for now.

To do this create a new "local" in the `root_local.tf`:
```
...
   s3_encryption_key_arn = local.environment == "intg" ? module.s3_external_kms_key.kms_key_arn : ""
...
```

Then use this local here instead of the arn directly:
```
...
   kms_key_id = local.s3_encryption_key_arn
...
```

TomJKing (commented):
Looks good. Couple of comments. Thanks

root.tf
vimleshtna: I have removed the encryption key, thanks

TomJKing (approved):
Looks good. Thanks

PR #373, opened 2023-10-04
Enable export bucket encryption on staging

TomJKing:
To be deployed when TRE and FCL ready for testing on staging


root_locals.tf
```diff
@@ -107,8 +107,8 @@ locals {
   tre_environment = local.environment == "intg" ? "int" : local.environment
 
   // apply s3 bucket encryption in intg only for now
```
thanhz, line 109: This comment should be updated or removed

thanhz (approved):
Looks good

PR #382, opened 2023-10-13
Akka licence available won't be using 'http4s'

TomJKing:


vimleshtna (approved):
Looks good. Thanks
I assume we have a clean-up task to remove http4s related codes

TomJKing:
@vimleshtna There is a separate task to clean up the http4s code.

PR #390, opened 2023-11-03
Pass s3 export bucket access roles as array

TomJKing:
Requires PR: https://github.com/nationalarchives/tdr-terraform-modules/pull/268

Further IAM roles will need access to the s3 export buckets

Pass these roles as an array instead of individually

### DIFF
--- a/root.tf
+++ b/root.tf
@@ -489,25 +489,25 @@ module "export_step_function" {
 }
 
 module "export_bucket" {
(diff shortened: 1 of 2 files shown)


vimleshtna (approved):
Looks good. Thanks

vimleshtna (approved):
Looks good. Thanks

PR #394, opened 2023-11-06
Increase cpu for 'staging' and 'prod' environments

TomJKing:
As part of the load testing for TDR an initial recommendation would be to increase the CPU for the Consignment API ECS task.

Increasing CPU to 1024 (1 vCPU) will potentially give the following advantages:
- Improve the overall stability of TDR when a transfer with a large number of files (1000+) is being processed with other transfers occurring in parallel. Not increasing the CPU is likely to see a degradation to judgment users when larger transfers are happening
- Improve the UI responsiveness for a transfer with a large number of files, eg user adding additional metadata etc. Some of the UI improvements and additional features being added (eg adding references) are likely to put further pressure on the API making this increase more likely to be needed regardless of trying to increase the number of files TDR can handle. Increase the maximum number of files per transfer TDR can take. This is of secondary benefit compared to the above points.

See details here: https://national-archives.atlassian.net/browse/TDR-726

### DIFF
--- a/modules/consignment-api/ecs.tf
+++ b/modules/consignment-api/ecs.tf
@@ -1,6 +1,9 @@
 locals {
   app_port           = 8080
   ecr_account_number = var.environment == "sbox" ? data.aws_caller_identity.current.account_id : data.aws_ssm_parameter.mgmt_account_number.value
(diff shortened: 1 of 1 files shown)


vimleshtna (approved):
Looks good. Thanks

PR #395, opened 2023-11-08
TDR-3319 Move some of the github roles/policies to the tdr-terraform-github

thanhz:


TomJKing (commented):
Looks good. Thanks. 

These changes will need to be tested and deployed in a specific order, to ensure overwriting doesn't happen:

1. The `tdr-terraform-environments` changes will need to be deployed first, this will remove the resources
2. The `tdr-terraform-github` changes should then be deployed, this will re-create the resources in each environment.

I suggest before merging the changes for both the `tdr-terraform-environments` and `tdr-terraform-github` do the deployments from your local machine to intg to make sure everything works OK. If it doesn't it can still be rolled back easily.

TomJKing (approved):
Looks good. Thanks

PR #406, opened 2024-01-08
Enable metadata validation library on staging

vimleshtna:


root_locals.tf
```diff
@@ -124,6 +124,6 @@ locals {
 
   //feature access blocks
   block_assign_file_references = local.environment == "intg" ? false : true
-  block_validation_library     = local.environment == "intg" ? false : true
+  block_validation_library     = (local.environment == "intg" || local.environment == "staging") ? false : true
```
TomJKing, line 127: Not that its that important but can change this to: `block_validation_library = local.environment == "prod" ? true : false`

Just a little bit simpler.

TomJKing (commented):
Looks good. Just a minor comment. Thanks

root_locals.tf
vimleshtna: Yeah, thanks.

TomJKing (approved):
Looks good. Thanks

PR #409, opened 2024-01-10
Roll back release of validation library on staging

TomJKing:
Metadata validation library has broken the entering of metadata


TomJKing:
Not needed

PR #413, opened 2024-01-22
TDR-3614- Update IAM role permissions to enable use of relevant KMS keys

annielh:


root_consignment_export.tf
```diff
@@ -49,6 +49,7 @@ module "consignment_export_task_policy" {
       titleEnvironment          = title(local.environment),
       aws_region                = local.region,
       account                   = data.aws_caller_identity.current.account_id,
+      //TODO: change this argument in tdr-terraform-modules to accept decryption keys array (or upgrade to a component from da-terraform-modules) and pass through internal bucket key in addition
```
TomJKing, line 52: @annielh Does migrating to the da-terraform-modules mean the entire ECS task is removed / re-created?

root_consignment_export.tf
```diff
@@ -45,11 +45,14 @@ module "consignment_export_task_policy" {
...
-      environment               = local.environment,
-      titleEnvironment          = title(local.environment),
-      aws_region                = local.region,
-      account                   = data.aws_caller_identity.current.account_id,
-      kms_export_bucket_key_arn = module.s3_external_kms_key.kms_key_arn
+      environment      = local.environment,
+      titleEnvironment = title(local.environment),
+      aws_region       = local.region,
+      account          = data.aws_caller_identity.current.account_id,
+      kms_export_bucket_key_arns = jsonencode([
```
TomJKing, line 52: Maybe rename `kms_export_bucket_key_arns` to reflect that its not just the export bucket keys any more: `kms_bucket_key_arns`

This will mean changing the corresponding code in the tdr-terraform-modules as well to have consistent naming.

TomJKing (commented):
Looks good. Thanks. Just one comment.

Once these changes are all ready I think it would be best to deploy them locally to intg and check everything works, before merging the changes into main. That way we can back them out quickly should something get broken.

TomJKing (approved):
Looks good. Thanks

TomJKing (approved):
Looks good. Thanks.

PR #424, opened 2024-02-06
TDR 3686- Encrypt internal s3 buckets with customer managed KMS key

annielh:
This also bumps `da-terraform-configurations` to pass through the role used by github for e2e testing to be passed through as a user role to the kms module (https://github.com/nationalarchives/da-terraform-configurations/pull/29).


TomJKing (approved):
Looks good. Thanks

PR #426, opened 2024-02-08
Tdr 3600 add permission to access s3 bucket

vimleshtna:


root_locals.tf
```diff
@@ -100,6 +100,7 @@ locals {
   runtime_java_11    = "java11"
 
   upload_files_cloudfront_dirty_bucket_name = "${var.project}-upload-files-cloudfront-dirty-${local.environment}"
+  draft_metadata_bucket_name                = "${var.project}-draft-metadata-${local.environment}"
```
TomJKing, line 103: Is there a reason to define the bucket name here? Its only used within the `root_draft_metadata.tf` I would define it as a `local` within the `rootf_draft_metadata.tf` unless its needed else where in the code.

TomJKing (commented):
Looks good. Just one minor comment.

Assume those permission work to allow the Lambda to read/write from the bucket.

root_locals.tf
vimleshtna: Yeah, we can define it within the `rootf_draft_metadata.tf`

TomJKing (approved):
Looks good. Thanks

PR #429, opened 2024-02-19
TDRE 54: Turn on encryption for upload bucket in intg

annielh:
This turns on encryption with our new customer managed KMS key for our upload bucket in s3, as well as upgrading submodules to support giving the necessary permissions to cloudfront to work with SSE-KMS.

It passes the distribution arn through to the s3 and kms modules to put in place the necessary permissions to enable use of the key from both sides.

It also passes the flag for the the encryption being enable to the cloudfront module, allowing it to use an OAC (original access control) rather than OAI (origin access identity) origin, as this is required for Cloudfront to work with SSE-KMS.

After releasing to all environments, we may want to return and remove the OAI origin altogether, but for the time being selection is driven by this new boolean.

---
### Test Plan
Deployed to intg; verified that put request to upload bucket still working. Also verified that the same request worked with encryption switched off, the legacy OAI target being selected as the cloudfront origin.

### DIFF
--- a/root_locals.tf
+++ b/root_locals.tf
@@ -120,7 +120,7 @@ locals {
   internal_bucket_key_enabled    = local.environment == "intg"
 
   // s3 upload bucket encryption
-  upload_dirty_s3_encryption_key_arn = "" // local.environment == "intg" ? module.s3_upload_kms_key.kms_key_arn : ""
+  upload_dirty_s3_encryption_key_arn = ""    // local.environment == "intg" ? module.s3_upload_kms_key.kms_key_arn : ""
   upload_dirty_bucket_key_enabled    = false // local.environment == "intg"
 
   // event bus hosted on tre environments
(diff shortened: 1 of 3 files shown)


Tom-Hallett (approved):
All good, thanks

vimleshtna (dismissed):
Looks good. Thanks

root_locals.tf
```diff
@@ -120,8 +120,8 @@ locals {
   internal_bucket_key_enabled    = local.environment == "intg"
 
   // s3 upload bucket encryption
-  upload_dirty_s3_encryption_key_arn = ""
-  upload_dirty_bucket_key_enabled    = false
+  upload_dirty_s3_encryption_key_arn = local.environment == "intg" ? module.s3_upload_kms_key.kms_key_arn : ""
+  upload_dirty_bucket_key_enabled    = true
```
TomJKing, line 124: Think this should be behind a ternary as well? Though should this be set to true regardless?

TomJKing (commented):
Looks good. Just one minor question

root_locals.tf
annielh: Thanks for the spot!

Tom-Hallett (approved):
Looks good. Thanks

PR #278, opened 2022-10-20
TDR-2391 Create new SQS queue

thanhz:
Create a new SQS queue
Created ssm parameter containing tre-out arn 

There will be a seperate pr to pass in an additional variable (sns_topic_arns) with the value from the SSM parameter once tre allows our SQS subscribe to tre-out SNS)

PR #448, opened 2024-04-10
TDRD-155 DDT access to TDR s3 standard export buckets

thanhz:
Will update submodule hash after https://github.com/nationalarchives/tdr-terraform-modules/pull/278 is merged


TomJKing (approved):
Looks good. Thanks

PR #453, opened 2024-04-30
Add a new export task.

MancunianSam:
There are a few changes here.

The step function has a new export task in a parallel block with the old
one. This calls the same task definition with different parameters.

The task definition has been updated with some new environment
variables.

The task definition task role has been updated to with new permissions.

There is a new SNS topic which receives messages when a file is exported
by the new process.
The permissions for this have been set to allow the DR2 account access
to subscribe to the topic. This does seem a bit permissive but it's what
AWS recommend https://docs.aws.amazon.com/sns/latest/dg/sns-send-message-to-sqs-cross-account.html#SendMessageToSQS.cross.account.queueowner
if we want the queue owner (DR2) to create the subscription which I
think we do.


root.tf
```diff
@@ -459,16 +464,20 @@ module "export_step_function" {
...
+    account_id                    = data.aws_caller_identity.current.account_id
+    environment                   = local.environment
+    security_groups               = jsonencode([module.consignment_export_ecs_security_group.security_group_id]),
+    subnet_ids                    = jsonencode(module.export_efs.private_subnets),
+    cluster_arn                   = module.consignment_export_ecs_task.cluster_arn,
+    task_arn                      = module.consignment_export_ecs_task.task_definition_arn,
+    task_name                     = "consignment-export",
+    sns_topic                     = module.notifications_topic.sns_arn,
+    platform_version              = "1.4.0"
+    max_attempts                  = 3,
```
annielh, line 476: @thanhz is taking this down to 2 in is #452, in case it doesn't flag in rebase process

annielh (dismissed):
LGTM

root.tf
```diff
@@ -484,6 +493,36 @@ module "export_step_function" {
   })
 }
 
+module "new_export_bucket" {
```
TomJKing, line 496: Maybe rename the module something a bit more meaningful such as: `flat_format_export_bucket`

root.tf
```diff
@@ -484,6 +493,36 @@ module "export_step_function" {
...
+  source            = "./da-terraform-modules/s3"
+  bucket_name       = "tdr-export-${local.environment}"
+  kms_key_arn       = module.s3_external_kms_key.kms_key_arn
+  common_tags       = local.common_tags
+  use_random_suffix = true
+}
+
+module "new_export_bucket_judgment" {
```
TomJKing, line 504: Maybe rename the module something a bit more meaningful such as: `flat_format_export_bucket_judgment`

root.tf
```diff
@@ -484,6 +493,36 @@ module "export_step_function" {
   })
 }
 
+module "new_export_bucket" {
+  source            = "./da-terraform-modules/s3"
+  bucket_name       = "tdr-export-${local.environment}"
```
TomJKing, line 498: The bucket name should be suffixed with a random id.

root.tf
```diff
@@ -484,6 +493,36 @@ module "export_step_function" {
...
+
+module "new_export_bucket_judgment" {
+  source            = "./da-terraform-modules/s3"
+  bucket_name       = "tdr-export-judgment-${local.environment}"
+  kms_key_arn       = module.s3_external_kms_key.kms_key_arn
+  common_tags       = local.common_tags
+  use_random_suffix = true
+}
+
+module "export_sns_notifications_topic" {
```
TomJKing, line 512: This topic should be made more generic so not just export notification can be sent to it.

Suggest changing:

- module name to: `external_sns_notifications_topic`
- `topic_name` local to: `local.external_notifications_topic_name`
- `sns_policy` template file name to: `.../sns/external_notifications_policy ...`

root.tf
```diff
@@ -493,6 +532,7 @@ module "export_bucket" {
   bucket_key_enabled    = true
   read_access_role_arns = local.standard_export_bucket_read_access_roles
   bucket_policy         = "export_bucket"
+
```
TomJKing, line 535: Unnecessary line break

root.tf
```diff
@@ -766,7 +806,8 @@ module "api_database_security_group" {
     { port = 5432, description = "Allow inbound access from database migrations", security_group_id = module.database_migrations.db_migration_security_group },
     { port = 5432, description = "Allow inbound access from create bastion users lambda", security_group_id = module.create_bastion_user_lambda.create_users_lambda_security_group_id[0] },
     { port = 5432, description = "Allow inbound access from create users lambda", security_group_id = module.create_db_users_lambda.create_users_lambda_security_group_id[0] },
-    { port = 5432, description = "Allow inbound access from backend checks", security_group_id = module.outbound_with_db_security_group.security_group_id }
+    { port = 5432, description = "Allow inbound access from backend checks", security_group_id = module.outbound_with_db_security_group.security_group_id },
+    { port = 5432, description = "Allow inbound access from export", security_group_id = module.consignment_export_ecs_security_group.security_group_id }
```
TomJKing, line 810: What is this for?

root_encryption.tf
```diff
@@ -21,6 +21,24 @@ module "s3_external_kms_key" {
...
 }
 
+module "sns_external_kms_key" {
+  source   = "./da-terraform-modules/kms"
+  key_name = "tdr-sns-external-kms-${local.environment}"
+  tags     = local.common_tags
+  default_policy_variables = {
+    user_roles = concat([
+      module.consignment_export_task_role.role.arn,
+    ], local.aws_sso_export_bucket_access_roles)
```
TomJKing, line 31: Don't think should use the `local.aws_sso_export_bucket_access_roles` as this is just access to the SNS topic encryption. That group of roles includes "export" which is for digital archivist to get access to the S3 buckets containing the bagit packages. Not sure that is relevant here.

root_locals.tf
```diff
@@ -26,6 +26,8 @@ locals {
 
   database_ca_cert_identifier = "rds-ca-rsa2048-g1"
 
+  export_notifications_topic_name = "tdr-external-notifications-${local.environment}"
```
TomJKing, line 29: Rename this local to: `external_notifications...`

templates/sns/export_notifications_policy.json.tpl
```diff
@@ -0,0 +1,24 @@
...
+      "Sid": "SNSAllowDR2ToSubscribe",
+      "Effect": "Allow",
+      "Principal": {
+        "AWS": "${dr2_account_number}"
+      },
+      "Action": "sns:Subscribe",
+      "Resource": "arn:aws:sns:${region}:${account_id}:${topic_name}"
+    }
+  ]
+}
```
TomJKing, line 24: Missing line break at end of file

templates/step_function/consignment_export_definition.json.tpl
```diff
@@ -1,19 +1,136 @@
 {
   "Comment": "A state machine to run the Fargate task to export the consignment",
-  "StartAt": "Run ECS task",
+  "StartAt": "Parallel",
```
TomJKing, line 3: You'll need to incorporate Thanh's changes to implement retry logic on ECS execution failures into the separate task definitions: https://github.com/nationalarchives/tdr-terraform-environments/pull/452

templates/step_function/consignment_export_definition.json.tpl
```diff
@@ -129,4 +216,4 @@
       "Type": "Fail"
     }
   }
-}
+}
```
TomJKing, line 219: Missing line break at end of file

templates/step_function/consignment_export_definition.json.tpl
```diff
@@ -1,19 +1,136 @@
 {
   "Comment": "A state machine to run the Fargate task to export the consignment",
```
TomJKing, line 2: Can you update the comment to reflect that the state machine now generates two types of export output.

templates/ecs_tasks/consignment_export.json.tpl
```diff
@@ -15,21 +19,25 @@
         "value": "${clean_bucket}"
       },
       {
-        "name": "OUTPUT_BUCKET",
```
TomJKing, line 18: Why are these bucket environment variables not needed any more?

TomJKing (commented):
Looks good. Mostly just minor comments and questions. Thanks

root.tf
MancunianSam: I think I've updated this since you wrote it but the module now has a `use_random_suffix` variable that adds it on to the name. It's there to stop breaking existing buckets. I've set it to true on these two so they'll get the random name.

root.tf
MancunianSam: Fair points, I've changed them all.

root.tf
MancunianSam: this is the security group for the database. It needs to allow the security group for the export task to access it as we're talking directly to the DB.

root_encryption.tf
MancunianSam: Ah ok, I wasn't sure so I guessed. I've taken it out.

templates/step_function/consignment_export_definition.json.tpl
MancunianSam: I did. They both have the `"ECS.AmazonECSException"` in the catch block and use the `max_attempts` variable. You might have written this before I did that now I think about it.

templates/step_function/consignment_export_definition.json.tpl
MancunianSam: I had Intellij configured to add these in and it seems to have stopped :frowning_face:

templates/ecs_tasks/consignment_export.json.tpl
MancunianSam: They're set in the step function. We're running the same task for both processes, the only difference being the run command and the export bucket so they're added in there instead. It reduces the amount of new stuff I had to create.

TomJKing (commented):
Looks good. Thanks

PR #454, opened 2024-05-07
TDRD 88- Evolved draft metadata step function and related infrastructure

annielh:
- Give antivirus lambda access to metadata bucket
- Create step function which calls antivirus with `metadata` scanType before calling the draft metadata validator lambda
  - Update api gateway definition to call this step function instead of the lambda directly
  - Create and include a role enabling the execution of the step function from api gateway
- If a virus is found, update the db by calling the consignment API directly from the step function
  - Set up an eventbridge connection to the consignment API, authorised via keycloak, to enable this 
  - Add a policy with the necessary permissions for this invocation and attach it to the step function role

**NB:** For now we set the `DraftMetadata` status to `Failed` if a virus is detected as it is one of the currently allowed values for a consignment status. We may want to consider adding a new `VirusDetected` status in the consignment api (looks like a pretty straightforward PR), although this would mean adding a status value that may only be applicable for one of the consignment stages. Happy to hear thoughts.

### DIFF
--- a/root_draft_metadata.tf
+++ b/root_draft_metadata.tf
@@ -166,7 +166,7 @@ resource "aws_iam_policy" "api_invoke_policy" {
           "secretsmanager:GetSecretValue",
           "secretsmanager:DescribeSecret"
         ]
-        Resource = "arn:aws:secretsmanager:*:*:secret:events!connection/*"
+        Resource = "arn:aws:secretsmanager:${local.region}:${var.tdr_account_number}:secret:events!connection/*"
       }
     ]
   })
--- a/root_draft_metadata.tf
+++ b/root_draft_metadata.tf
@@ -82,70 +82,70 @@ resource "aws_events_connection" "consignment_api_connection" {
 }
 
 module "draft_metadata_checks" {
-  source = "./da-terraform-modules/sfn"
+  source             = "./da-terraform-modules/sfn"
   step_function_name = "TDRMetadataChecks${title(local.environment)}"
   step_function_definition = jsonencode(
     {
-      "Comment": "Run antivirus checks on metadata, update DB if positive, else trigger metadata validation",
-      "StartAt": "RunAntivirusLambda",
-      "States": {
-        "CallCheckLambda": {
-          "Type": "Task",
-          "Resource": module.yara_av_v2.lambda_arn
-          "Parameters": {
-            "consignmentId.$": "$.consignmentId",
-            "fileId.$": "draft-metadata.csv",
-            "scanType.$": "metadata"
+      "Comment" : "Run antivirus checks on metadata, update DB if positive, else trigger metadata validation",
+      "StartAt" : "RunAntivirusLambda",
+      "States" : {
+        "CallCheckLambda" : {
+          "Type" : "Task",
+          "Resource" : module.yara_av_v2.lambda_arn
+          "Parameters" : {
+            "consignmentId.$" : "$.consignmentId",
+            "fileId.$" : "draft-metadata.csv",
+            "scanType.$" : "metadata"
           },
-          "Next": "CheckAntivirusResults"
+          "Next" : "CheckAntivirusResults"
         },
-        "CheckAntivirusResults": {
-          "Type": "Choice",
-          "Choices": [
+        "CheckAntivirusResults" : {
+          "Type" : "Choice",
+          "Choices" : [
             {
-              "Variable": "$.result",
-              "StringEquals": "",
-              "Next": "ValidateMetadataLambda"
+              "Variable" : "$.result",
+              "StringEquals" : "",
+              "Next" : "ValidateMetadataLambda"
             },
             {
-              "Not": {
-                "Variable": "$.result",
-                "StringEquals": ""
+              "Not" : {
+                "Variable" : "$.result",
+                "StringEquals" : ""
               },
-              "Next": "PrepareVirusDetectedQueryParams"
+              "Next" : "PrepareVirusDetectedQueryParams"
             }
           ],
-          "Default": "CallValidateMetadataLambda"
+          "Default" : "CallValidateMetadataLambda"
         },
-        "PrepareVirusDetectedQueryParams": {
-          "Type": "Pass",
-          "ResultPath": "$.statusUpdate",
-          "Parameters": {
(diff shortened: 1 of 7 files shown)


root_draft_metadata.tf
```diff
@@ -46,9 +47,212 @@ module "draft_metadata_api_gateway" {
...
+    ]
+  })
+}
+
+module "draft_metadata_checks" {
+  source             = "./da-terraform-modules/sfn"
+  step_function_name = "TDRMetadataChecks${title(local.environment)}"
+  step_function_definition = jsonencode(
+    {
+      "Comment" : "Run antivirus checks on metadata, update DB if positive, else trigger metadata validation",
```
TomJKing, line 180: Maybe put `DB` out in full, ie "database"

root_draft_metadata.tf
```diff
@@ -46,9 +47,212 @@ module "draft_metadata_api_gateway" {
...
+      "StartAt" : "RunAntivirusLambda",
+      "States" : {
+        "RunAntivirusLambda" : {
+          "Type" : "Task",
+          "Resource" : module.yara_av_v2.lambda_arn
+          "Parameters" : {
+            "consignmentId.$" : "$.consignmentId",
+            "fileId" : "draft-metadata.csv",
```
TomJKing, line 188: This probably should not be hard coded, in case the name is changed for the CSV file etc. Can it be passed in from the step function request?

root_draft_metadata.tf
```diff
@@ -46,9 +47,212 @@ module "draft_metadata_api_gateway" {
...
+        "PrepareVirusDetectedQueryParams" : {
+          "Type" : "Pass",
+          "ResultPath" : "$.statusUpdate",
+          "Parameters" : {
+            "query" : "mutation updateConsignmentStatus($updateConsignmentStatusInput: ConsignmentStatusInput!) { updateConsignmentStatus(updateConsignmentStatusInput: $updateConsignmentStatusInput) }",
+            "variables" : {
+              "updateConsignmentStatusInput" : {
+                "consignmentId.$" : "$.consignmentId",
+                "statusType" : "DraftMetadata",
+                "statusValue" : "Failed"
```
TomJKing, line 221: I would have this as `CompletedWithIssues` The `Failed` status means that there is an actual failure with the code, ie the step function fails to run, the Lambda crashes etc. Here the code executes correctly and is doing the right thing, so it completed but found a problem with the input.

There is needs to be a re-appraisal of the statuses for draft metadata and more generally though.

root_draft_metadata.tf
```diff
@@ -46,9 +47,212 @@ module "draft_metadata_api_gateway" {
   }]
 }
 
+resource "aws_iam_role" "draft_metadata_api_gateway_execution_role" {
+  name = "TDRMetadataChecksAPIGatewayExecutionRole${title(local.environment)}"
+
+  assume_role_policy = jsonencode({
```
TomJKing, line 53: TDR uses json files instead of directly having policies defined in the `.tf` files themselves.

root_draft_metadata.tf
```diff
@@ -46,9 +47,212 @@ module "draft_metadata_api_gateway" {
...
+}
+
 module "draft_metadata_bucket" {
   source      = "./da-terraform-modules/s3"
   bucket_name = local.draft_metadata_s3_bucket_name
   common_tags = local.common_tags
   kms_key_arn = module.s3_internal_kms_key.kms_key_arn
 }
+
+data "aws_ssm_parameter" "backend_checks_keycloak_secret" {
```
TomJKing, line 88: There is a story to create a new Keycloak client for the draft metadata, so this will need to be changed as part of that task: https://national-archives.atlassian.net/browse/TDRD-181

root_draft_metadata.tf
```diff
@@ -46,9 +47,212 @@ module "draft_metadata_api_gateway" {
...
+      }
+    }
+  }
+}
+
+resource "aws_iam_policy" "draft_metadata_checks_policy" {
+  name        = "TDRMetadataChecksPolicy${title(local.environment)}"
+  description = "Policy to allow necessary lambda executions from step function"
+
+  policy = jsonencode({
```
TomJKing, line 122: TDR uses json files instead of directly having policies defined in the `.tf` files themselves.

TomJKing (commented):
Looks good. Just some minor comments. Thanks

root_draft_metadata.tf
annielh: :+1:

root_draft_metadata.tf
```diff
@@ -46,9 +47,158 @@ module "draft_metadata_api_gateway" {
...
+  policy = templatefile("./templates/iam_policy/invoke_lambda_policy.json.tpl", {
+    resources = jsonencode([
+      module.yara_av_v2.lambda_arn,
+      module.draft_metadata_validator_lambda.lambda_arn
+    ])
+  })
+}
+
+resource "aws_iam_policy" "api_invoke_policy" {
+  name = "TDRAPIInvokePolicy${title(local.environment)}"
```
TomJKing, line 110: Should rename this policy to match what it is doing: `TDRAPIThirdPartyInvokePolicy...`

root_draft_metadata.tf
```diff
@@ -46,9 +47,158 @@ module "draft_metadata_api_gateway" {
...
+    connection_arn    = aws_cloudwatch_event_connection.consignment_api_connection.arn
+    api_url           = module.consignment_api.api_url
+    step_function_arn = module.draft_metadata_checks.step_function_arn
+  })
+}
+
+module "draft_metadata_checks" {
+  source             = "./da-terraform-modules/sfn"
+  step_function_name = "TDRMetadataChecks${title(local.environment)}"
+  step_function_definition = jsonencode(
```
TomJKing, line 124: The step function definition should go in a `.json.tpl` file. See here: https://github.com/nationalarchives/tdr-terraform-environments/tree/master/templates/step_function

TomJKing (commented):
Looks good. Just a couple of other things I've noticed, apologies. Thanks

root_draft_metadata.tf
annielh: This particular instance of the policy (with the `api_url` being our own consignment API) isn't really invoking a third party api so had erred on leaving it out as it feels slightly misleading. Would we want anything using that template to have the same name? If anything if we are going to rename I would lean towards `TDRConsignmentAPIInvokePolicy...`.  Let me know what you think. Will need a change in the backend too.

TomJKing: Fair point

PR #458, opened 2024-06-05
Update `tdr-configurations` submodule hash

thanhz:
Contains update to ssoSessiontimeout in the realm.json


thanhz:
Merged as part of another pr already

PR #459, opened 2024-06-05
Set ALB logging bucket encryption to default

TomJKing:
For ALB logging to work the s3 bucket encryption should be set to the default encryption: 'Amazon S3-managed keys (SSE-S3)'

The existing s3 bucket has got into a state where was being encrypted by a non-existent customer managed key preventing any new ALB to log to it

The objects stored in the s3 bucket whilst using the non-existent customer managed key are still encrypted using the SSE-S3 encryption. So the bucket encryption can be reset to this


thanhz (approved):
Looks good to me :+1:

PR #462, opened 2024-06-07
IAM roles and policies for transfer service ecs task

TomJKing:
Add IAM roles and policies to allow set up of ECR repository permissions for new ECS task image


templates/iam_policy/transfer_service_ecs_task_policy.json.tpl
```diff
@@ -0,0 +1,17 @@
...
+  "Statement": [
+   {
+         "Effect": "Allow",
+         "Action": [
+           "kms:Decrypt",
+           "kms:GenerateDataKey",
+           "sns:Publish"
+         ],
+         "Resource": [
+           "arn:aws:kms:eu-west-2:229554778675:key/b05401ed-cc76-46f6-bf2c-76cb0a859542",
```
annielh, line 12: Do we need this arn hard coded here? Are we able to put it in `da-terraform-configurations`?

PR #470, opened 2024-06-13
Transfer service ecs task

TomJKing:
Will replace with `da-terraform-modules` when available.

### DIFF
--- a/root_transfer_service.tf
+++ b/root_transfer_service.tf
@@ -60,7 +60,6 @@ module "transfer_service_route53" {
   common_tags           = local.common_tags
   environment_full_name = local.environment_full_name
   project               = "tdr"
-  cname_record_name     = "transfer-service"
   a_record_name         = "transfer-service"
   alb_dns_name          = module.transfer_service_tdr_alb[0].alb_dns_name
   alb_zone_id           = module.transfer_service_tdr_alb[0].alb_zone_id
--- a/root_transfer_service.tf
+++ b/root_transfer_service.tf
@@ -1,0 +1,145 @@
+locals {
+  // Apply to intg only intially
+  transfer_service_count = local.environment == "intg" ? 1 : 0
+}
+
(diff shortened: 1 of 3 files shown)


Tom-Hallett (dismissed):
All looks fine

root_transfer_service.tf
```diff
@@ -42,3 +43,106 @@ module "transfer_service_task_policy" {
...
+  container_name               = "transfer-service"
+  cpu                          = 512
+  environment                  = local.environment
+  execution_role               = module.transfer_service_execution_role[0].role_arn
+  load_balancer_container_port = 8080
+  memory                       = 1024
+  private_subnets              = module.shared_vpc.private_backend_checks_subnets
+  security_groups              = [module.transfer_service_ecs_security_group[0].security_group_id]
+  service_name                 = "transfer_service_${local.environment}"
+  task_family_name             = "transfer_service-${local.environment}"
```
annielh, line 146: Nit- choose `_` or `-` for both

annielh (approved):
All looks fine, tiny tiny comment.

I guess my only question would be how much work would it be to take the opportunity to add the necessary modules to `da-terraform-modules` now? Do we have a particular point where we're expecting them to become available through our or other teams work if we don't add them now / do we have a placeholder for the work?

root_transfer_service.tf
TomJKing: Good spot

TomJKing:
> _I guess my only question would be how much work would it be to take the opportunity to add the necessary modules to da-terraform-modules now?_
@annielh I will need to add the new modules, I've started this work, but it will take a while for me to do it all. I want to get this ECS task stood up for a longer period of time to test stuff out in preparation to the SharePoint work. Which is why I'm using the tdr modules for now. Does that make sense?

annielh (approved):

PR #476, opened 2024-07-03
TDRD-227 - Add "blockMetadataReview" feature access block for frontend

vimleshtna:


TomJKing (approved):
Looks good, provided only blocking the feature in Prod. Thanks

PR #478, opened 2024-07-05
Allow lambda to use ecr

ian-hoyle:
This is just to see if the lambda using the ecr can be created
If it works will delete the old draft-metadata validator and replace with this

### DIFF
--- a/root_draft_metadata_ecr.tf
+++ b/root_draft_metadata_ecr.tf
@@ -1,0 +1,31 @@
+
+#
+module "draft_metadata_validator_lambda_ecr" {
+  source          = "./da-terraform-modules/lambda"
+  function_name   = "tdr-draft-metadata-validator-2-${local.environment}"
+  handler         = "uk.gov.nationalarchives.draftmetadatavalidator.Lambda::handleRequest"
+  tags            = local.common_tags
+  use_image       = true
+  image_url       = "arn:aws:ecr:eu-west-2:${data.aws_ssm_parameter.mgmt_account_number.value}:repository/draft-metadata-validator"
+  timeout_seconds = 120
+  memory_size     = 1024
+  policies = {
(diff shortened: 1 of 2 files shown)


root_draft_metadata_ecr.tf
```diff
@@ -0,0 +1,25 @@
+module "draft_metadata_validator_lambda_ecr" {
+  source          = "./da-terraform-modules/lambda"
+  function_name   = "tdr-draft-metadata-validator-2-${local.environment}"
```
TomJKing, line 3: Could you rename the function: `...-validator-v2-...` Its the pattern used for other revised Lambdas on TDR eg `tdr-yara-av-v2-intg`

root_draft_metadata_ecr.tf
```diff
@@ -0,0 +1,25 @@
...
+    "TDRDraftMetadataValidatorLambdaPolicyECR${title(local.environment)}" = templatefile("./templates/iam_policy/draft_metadata_validator_lambda.json.tpl", {
+      account_id     = var.tdr_account_number
+      environment    = local.environment
+      parameter_name = local.keycloak_backend_checks_secret_name
+      bucket_name    = local.draft_metadata_s3_bucket_name
+      kms_key_arn    = module.s3_internal_kms_key.kms_key_arn
+    })
+  }
+  plaintext_env_vars = {
+    API_URL            = "graphql"
```
TomJKing, line 20: In the existing Lambda implementation the `API_URL` is set to the full Consignment API URL value.

In the Terraform it is set as: `"${module.consignment_api.api_url}/graphql"`

TomJKing (commented):
Looks good. Just a couple of very minor comments. Thanks.

root_draft_metadata_ecr.tf
ian-hoyle: changed

TomJKing (approved):
Looks good. Thanks

PR #479, opened 2024-07-09
Add file checks total timeouts as per the backend checks timout for the frontend

vimleshtna:


thanhz (commented):
Looks good, I think you have to update the `da-terraform-modules` again as there are new changes

vimleshtna:
> Looks good, I think you have to update the `da-terraform-modules` again as there are new changes

Good catch. Yeah, it might break. I'm gonna create a new branch with my changes only. Thanks

PR #481, opened 2024-07-09
Update tdr-configurations to include new ayr ip whitelisting

anthonyhashemi:
Update tdr-configurations to include new ayr ip whitelisting:
https://github.com/nationalarchives/tdr-configurations/pull/181


vimleshtna (approved):
Looks good. Thanks

PR #486, opened 2024-07-12
allow lambda to use ecr image

ian-hoyle:


root_draft_metadata_ecr.tf
```diff
@@ -0,0 +1,26 @@
+module "draft_metadata_validator_lambda_ecr" {
+  source          = "./da-terraform-modules/lambda"
+  function_name   = "tdr-draft-metadata-validator-v2-${local.environment}"
+  handler         = "uk.gov.nationalarchives.draftmetadatavalidator.Lambda::handleRequest"
+  tags            = local.common_tags
+  use_image       = true
+  image_url       = "${data.aws_ssm_parameter.mgmt_account_number.value}.dkr.ecr.eu-west-2.amazonaws.com/draft-metadata-validator:v41"
```
TomJKing, line 7: The `v41` should not be hard coded. Should use the environment tag, ie: `... draft-metadata-validator:${local.environment}`

templates/iam_policy/draft_metadata_validator_lambda_ecr.json.tpl
```diff
@@ -0,0 +1,51 @@
...
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": [
+        "logs:CreateLogStream",
+        "logs:PutLogEvents"
+      ],
+      "Resource": [
+        "arn:aws:logs:eu-west-2:${account_id}:log-group:/aws/lambda/tdr-draft-metadata-validator-${environment}",
```
TomJKing, line 11: A new log group for the new lambda will be created via the `da-terraform-modules`: https://github.com/nationalarchives/da-terraform-modules/blob/main/lambda/main.tf#L83

Therefore the IAM policy will need to be changed to reflect this: `arn:aws:logs:eu-west-2:${account_id}:log-group:/aws/lambda/tdr-draft-metadata-validator-v2-${environment}`

TomJKing (commented):
Looks good. Couple of minor comments. Thanks

root_draft_metadata_ecr.tf
ian-hoyle: I have changed it. Had hard coded the value to simplify release.
I am creating another github workflow to allow environment tagging to make it simpler

templates/iam_policy/draft_metadata_validator_lambda_ecr.json.tpl
ian-hoyle: I was going to use the same log group - the new lambda will have the same name tdr-draft-metadata-validator-${environment}. terraform destroys the old lambda and creates the new one with the same name but does nothing with the log group

TomJKing: At the moment the code in the PR has the functioned name as: `tdr-draft-metadata-validator-v2-${local.environment}` 

Running the Terraform plan a new lambda and CloudWatch log group tare to be to be created with the `v2` name. In which case the permissions in this policy won't work are they are reference the incorrect log group.

Plan snippets:
Screenshot from 2024-07-19 12-40-59
Screenshot from 2024-07-19 12-40-33

TomJKing (commented):
Additional comment. Thanks

PR #494, opened 2024-08-01
Deploy transfer service to staging

TomJKing:
SharePoint App will be connecting to staging for testing to avoid ip restrictions on intg


vimleshtna (approved):
Looks good. Thanks

PR #501, opened 2024-08-07
DR2-1842 Add copy lambda role to export bucket.

MancunianSam:
This should allow our lambda which copies files from the TDR export
bucket to our bucket to work.

I've only added this for the export bucket as we're not using the
judgment bucket.


root_locals.tf
```diff
@@ -114,6 +114,8 @@ locals {
 
   talend_export_role_arn = module.talend_configuration.terraform_config[local.environment]["remote_engine_instance_profile_role"]
 
+  dr2_copy_files_role = "arn:aws:iam::${module.dr2_configuration.account_numbers[local.environment]}:role/${local.environment}-dr2-copy-files-from-tdr-role"
```
TomJKing, line 117: It would be better for none of the arn to be hardcoded in case the name of the lambda changes, or is replaced by a different one. It just decouples TDR from any code changes on the DR2 side as much as possible.

So something like this: `arn:aws:iam::${module.dr2_configuration.account_numbers[local.environment]}:role/${module.dr2_configuration.copy_from_tdr_lambda_role}`

TomJKing (commented):
Looks good. Just one comment. Thanks

root_locals.tf
MancunianSam: I've added the whole arn to da-terraform-configurations and just referenced the whole thing in here.

TomJKing (approved):
Looks good. Thanks.

PR #503, opened 2024-08-08
Add dr2 role to external KMS key.

MancunianSam:
We need to copy the files from the new export bucket. The bucket policy
has this role in it but the bucket is encrypted with this key and I need
permission to decrypt.


TomJKing (approved):
Looks good. Thanks

PR #513, opened 2024-09-18
give draft_metadata front end policy permission to getObject 

ian-hoyle:
The front end needs to download the draft-metadata-errors.json file


vimleshtna (approved):
Looks good. Thanks

PR #519, opened 2024-10-02
Tdrd 532 ba iam permissions

TomJKing:
To support end to end testing of the Digital Archving pipeline provide a consistently named IAM policy that allows access to the relevant environments and AWS resources

These commonly named policies can be used to manage access across the services. See here for further details: https://github.com/nationalarchives/ds-infrastructure-identity/blob/main/README.md#managing-user-access-to-aws


thanhz (approved):
Looks okay to me :+1:

PR #520, opened 2024-10-08
Allow for SharePoint urls in origin

TomJKing:
SharePoint makes requests for signed cookies via the '/cookies' endpoint.

Ensure that the correct origin, ie the SharePoint site is return in the pre-flight request for the cookie to prevent CORS error on the client side

Ovveride the default origin if the origin in the pre-flight request belongs to the SharePoint domain, ie 'sharepoint.com'.

See details here regarding CORS and pre-flight requests: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#preflighted_requests

### DIFF
--- a/templates/api_gateway/sign_cookies_api.json.tpl
+++ b/templates/api_gateway/sign_cookies_api.json.tpl
@@ -79,7 +79,7 @@
                 "method.response.header.Access-Control-Allow-Origin" : "'${upload_cors_urls}'"
               },
               "responseTemplates" : {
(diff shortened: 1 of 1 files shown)


thanhz (commented):
I might be misunderstanding what this is for but but Just wondering why we don't enable CORS support on the API Gateway and then enter allowed origins that way? I might be confused, this file might be already doing that but I noticed cors support can be enabled on api gateway. 

image

TomJKing:
@thanhz Not sure. I think they do the same thing, but I'm a bit reluctant to change this at the moment as my knowledge is pretty limited on how this works.

thanhz (approved):

PR #531, opened 2024-10-25
TDRD-588: increase resources for file format lambda so works with 2GB file

MarkKingsbury:
2GB files failing again today.  Suspect that miss-reading logs with AV warnings as actual failures and root cause if that timeouts in this third “file format” lamda in the map.

Seems like partial download happens in each of 5 x 1 min timeout and that therefore it sometimes works on the sixth try.  or do the three lambda share a tmp directory and it has arrived from one of the other lambda in the map that have 5 min timouts? 

Tested with these extra resources and works.  Probably need to revisit how download works and storage, will discuss next week and maybe make card. And more testing.


TomJKing (approved):
Looks good. Thanks

PR #547, opened 2024-11-27
Tdrd 539 add security header

Tom-Hallett:


TomJKing (approved):
Looks good. Thanks.

Not sure the header also needs to go on the preflight `OPTIONS` method as well. 

Try this first. Should become apparent if it also needs to go on the `OPTIONS` as well.

PR #561, opened 2024-12-19
use latest da-terraform-modules, tdr-terraform-modules, domain = "vpc"

ian-hoyle:
Replaced vpc = true with domain = "vpc"
The domain argument is the current way to specify that the Elastic IP is for use in a VPC. 
overwrite attribute has been removed to remove warnings


TomJKing (approved):
Looks good. Thanks

PR #574, opened 2025-01-13
TDRD-22 - Add "condition" to the trust polices

vimleshtna:
Required: https://github.com/nationalarchives/tdr-terraform-modules/pull/305


modules/database-migrations/templates/migration_lambda.json.tpl
```diff
@@ -32,7 +42,12 @@
...
-      "Resource": ["*"]
+      "Resource": ["*"],
+      "Condition": {
+        "StringEquals": {
+          "AWS:SourceAccount": "${account_id}"
+        }
+      }
     }
   ]
 }
```
TomJKing, line 53: Missing line break at end of file

TomJKing (commented):
Looks good. Thanks. One very minor comment.

Would suggest testing this before merging to main, by doing an apply from the branch, so can roll back if necessary.

TomJKing (approved):
Looks good. Thanks.

PR #579, opened 2025-01-17
Tdrd 339 automated or semi automated workflow for blocking i ps with malicious activity

ian-hoyle:


root_locals.tf
```diff
@@ -60,6 +60,8 @@ locals {
 
   ip_allowlist = concat(local.developer_ip_list, local.trusted_ip_list)
 
+  ip_blocked_list = module.tdr_configuration.terraform_config["ip_blocked_list"]
```
TomJKing, line 63: Bit confused. From this PR: https://github.com/nationalarchives/da-terraform-configurations/pull/47 these are in the `da-terraform_configurations` not the `tdr_configuration`.

TomJKing (commented):
Just one query. Thanks.

root_locals.tf
ian-hoyle: The source for tdr_configuration is da-terraform-configurations
```
module "tdr_configuration" {
  source  = "./da-terraform-configurations"
  project = "tdr"
}
```

TomJKing: Didn't notice that.

TomJKing (approved):
Looks good. Thanks

PR #581, opened 2025-01-23
Adding Slack Webhook variable

Tom-Hallett:


root.tf
```diff
@@ -576,6 +576,7 @@ module "notification_lambda" {
   standard_export_s3_bucket_name = module.export_bucket.s3_bucket_name
   da_event_bus_arn               = local.da_event_bus_arn
   da_event_bus_kms_key_arn       = local.da_event_bus_kms_key
+  slack_bau_webhook              = module.bau_slack_channel_ssm_parameters.params[local.slack_bau_webhook].value
```
TomJKing, line 579: This doesn't need to be passed into the module. 

This is what is causing the error, as the variable `slack_bau_webhook` doesn't exist in the module: https://github.com/nationalarchives/tdr-terraform-modules/blob/master/lambda/variables.tf

You're already referencing the ssm parameter you created within the lambda terraform here to get its value: https://github.com/nationalarchives/tdr-terraform-modules/blob/master/lambda/notifications.tf#L96

TomJKing (commented):
Looks good. Thanks. Hope my comments make sense

TomJKing (approved):
Looks good. Thanks

PR #590, opened 2025-02-04
Remove OpenTelemetry

TomJKing:
OpenTelemetry metrics have caused a significant jump in CloudWatch cost

The cause is unclear

Removing until investigation and mitigation can be conducted


TomJKing:
Closing for now.

PR #591, opened 2025-02-04
Use dynamic lifecycle rules

TomJKing:
s3 modules sets rule by default, this causes conflict if create rule resource separately from the s3 module

Dynamically implement rules to avoid this


Tom-Hallett (approved):
Looks good, thanks

PR #608, opened 2025-03-04
TDRD-721 - Add Wiz IAM role to non-Prod TDR CMKs to allow scanning

vimleshtna:
Allow Wiz IAM role for following TDR CMKs:

- tdr-s3-external-kms-{intg/staging}
- tdr-sns-external-kms-{intg/staging}


TomJKing (approved):
Looks good. Thanks

PR #619, opened 2025-03-28
Nojira aws backups prod

TomJKing:


templates/s3/allow_read_access.json.tpl
```diff
@@ -2,6 +2,25 @@
   "Id": "allow-read-access-${bucket_name}",
   "Version": "2012-10-17",
   "Statement": [
+  %{ if aws_backup_local_role != "" }
```
annielh, line 5: I feel like it would be nicer having an array for the backup role and adding the array straight in as the principle, instead of the check on the empty string/logic in a template. But :heavy_check_mark: anyway if this needs deploying sharpish

PR #626, opened 2025-05-12
Adding SQS queue as External Event source

Tom-Hallett:


thanhz (approved):
Looks okay to me. Assuming you've already tried terraform plan?

PR #635, opened 2025-06-10
Draft metadata Lambda split step function update

ian-hoyle:
This PR is to add stubs for handling the response from the metadata validator lambda. It will handle the current response from the lambda and stubbed for handling the updated message form the lamba  that will be in a format similar to {
  "consignmentId": "uuid-value",
  "validationTime": "0.123 seconds",
  "validationLbraryVersion": "v2.3.1"
  "validationStatus": "success|failed",
  "errorFileUrl": "s3://bucket/consignment-id/error.json" // only when validation fails
}

### DIFF
--- a/root_draft_metadata.tf
+++ b/root_draft_metadata.tf
@@ -191,7 +191,6 @@ module "draft_metadata_checks" {
     antivirus_lambda_arn           = module.yara_av_v2.lambda_arn,
     consignment_api_url            = module.consignment_api.api_url,
     consignment_api_connection_arn = aws_cloudwatch_event_connection.consignment_api_connection.arn,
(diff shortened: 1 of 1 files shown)


ian-hoyle:
> This PR is to add stubs for handling the response from the metadata validator lambda. It will handle the current response from the lambda and stubbed for handling the updated message form the lamba that will be in a format similar to {
> "consignmentId": "uuid-value",
> "validationTime": "0.123 seconds",
> "validationLbraryVersion": "v2.3.1"
> "validationStatus": "success|failed",
> "errorFileUrl": "s3://bucket/consignment-id/error.json" // only when validation fails
> }

eaft

PR #638, opened 2025-06-12
fix: Confused deputy for flow logs

ben-cerium:
This is good to go.  Handcranked the changes it makes into int and plan shows no changes so safe to merge.

### DIFF
--- a/modules/shared-vpc/variables.tf
+++ b/modules/shared-vpc/variables.tf
@@ -2,8 +2,8 @@ variable "environment" {}
 variable "common_tags" {}
 variable "az_count" {}
 variable "database_availability_zones" {}
-variable "cloudwatch_rentention_period" {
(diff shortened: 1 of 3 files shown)


modules/shared-vpc/variables.tf
```diff
@@ -1,4 +1,9 @@
 variable "environment" {}
 variable "common_tags" {}
 variable "az_count" {}
-variable "database_availability_zones" {}
\ No newline at end of file
+variable "database_availability_zones" {}
+variable "cloudwatch_retention_period" {
```
TomJKing, line 5: Could you append `_days` to the variable name so it clear what the "measurement" is to calling clients: `cloudwatch_retention_period_days`

TomJKing (commented):
Looks good. Just one minor comment. Thanks

TomJKing (approved):
Looks good. Thanks

PR #639, opened 2025-06-17
Tdrd 699 add dr2 subscription

Tom-Hallett:


root_external_event_handling.tf
```diff
@@ -3,20 +3,35 @@ locals {
...
+    sqs_name         = local.sqs_name,
+    dr2_account_id   = module.dr2_configuration.account_numbers[local.environment],
+    dr2_ingest_topic = local.dr2_ingest_topic
+  })
   encryption_type    = "kms"
   kms_key_id         = module.encryption_key.kms_key_arn
   visibility_timeout = 6 * local.lambda_timeout
 }
 
+resource "aws_sns_topic_subscription" "dr2_ingest_complete" {
```
TomJKing, line 28: I would rename this: `dr2_ingest` It may or may not "complete" so best to keep it a bit less specific.

root_external_event_handling.tf
```diff
@@ -3,20 +3,35 @@ locals {
   event_handling_count                 = local.environment == "intg" ? 1 : 0
   sqs_name                             = "tdr-external-event-handling-sqs-${local.environment}"
   external_event_handler_function_name = "tdr-external-events-handler-${local.environment}"
+  dr2_ingest_topic                     = "${local.environment}-dr2-notifications"
```
TomJKing, line 6: Do the `d2` environments map to the TDR ones, eg intg / staging / prod? Using the `local.environment` here may cause problems if they don't as it refers to the TDR ones.

May be safer to get DR2 to add the topic to the `da-terraform-configurations` repo for TDR to read, rather than construct it within TDR's Terraform.

TomJKing (commented):
Looks good. Thanks. Couple of minor comments.

root_external_event_handling.tf
Tom-Hallett: DR2 environments match TDRs (according to da-terraform-configurations) but it's a fair point as other projects don't. Will discuss potential solutions with you.

PR #646, opened 2025-06-26
fix: Revert trust policy which is questionable now and should need be needed

ben-cerium:


ben-cerium:
In INT and tested https://github.com/nationalarchives/tdr-e2e-tests/actions/runs/15904333894

PR #675, opened 2025-08-20
Tdrd 932 replace antivirus lambda step

Tom-Hallett:


TomJKing (commented):
Looks good. Just some minor suggestions. Thanks

PR #683, opened 2025-08-28
Use new role name

MancunianSam:
We've updated the role name and the variable it's stored against, this
should now work.


Tom-Hallett (approved):
LGTM

PR #697, opened 2025-09-29
remove fab dry_run

ian-hoyle:
leave env variable in so dry run can be easily run if want to see how many users might be disabled. set to false 
Implement to be scheduled to run every 30 days, and disabling users who have not user TDR for 2 years.
Phase 1: 2 years / 730 days


vimleshtna (approved):
Looks good. Thanks

PR #858, opened 2026-04-09
Tdrd 1221/remove share vpc private vpc output

ben-cerium:
https://github.com/nationalarchives/tdr-e2e-tests/actions/runs/24189713405 tests ok

private subnet should be empty


TomJKing (approved):
Looks good. Thanks

PR #859, opened 2026-04-09
stop notification failure blowing up

ian-hoyle:
If the call to the notify lambda failed the error was not handled so step function failed. Fixed the notify lambda so it won't error and placed catch so even if error SF will complete.


TomJKing (approved):
Looks good. Thanks

PR #880, opened 2026-04-28
Disable otel collector on intg

TomJKing:
Wiz has identified several security issues with the otel image, which are awaitig resolution

Disabling across all enviroments until some time can be spent on otel/x-ray logging more generally


vimleshtna (approved):
Looks good. Thanks

PR #887, opened 2026-05-07
TDRD-939 - Add a new backend checks V2 step function to run the new file checks lambda

vimleshtna:


TomJKing (commented):
Looks good. Thanks

How are you going to control which version of step function to use on an environment, by environment basis?

vimleshtna:
> Looks good. Thanks
> 
> How are you going to control which version of step function to use on an environment, by environment basis?

Yeah, good point. Sorry, completely forgot about this. I have added a flag to disable only for prod.

TomJKing (approved):
Looks good. Thanks

PR #898, opened 2026-05-14
Use general deletion policy for export bucket

TomJKing:
'Mock' series have ingest message switched off so object are not getting tagged for deletion in lower environments

To manage bucket size do a general clean up all objects


vimleshtna (approved):
Looks good. Thanks

PR #917, opened 2026-06-05
TDRD-1461: increase front end memory 

MarkKingsbury:
- to allow for metadata download of 10k full metadata.
- previously only increased on staging, now do same on prod
- intg remains low, not intended to encounter these capacity levels.


TomJKing (approved):
Looks good. Thanks

PR #918, opened 2026-06-05
TDRD-1539: add retry condition in step functions

MarkKingsbury:
- whilst capacity testing encountered this `Events.ConnectionResource.AuthInProgress` in metadata checks step function so adding as additional retry alongside `Events.ConnectionResource.ConcurrentModification`
-  adding to two other step functions that share the `Events.ConnectionResource.ConcurrentModification`


TomJKing (approved):
Looks good. Thanks

PR #920, opened 2026-06-08
TDRD-1503 removed all mentions of bastion in root.tf

EnyoHA:


TomJKing:
@EnyoHA Your changes looks fine. Unfortunately I've noticed there is also a Lambda that needs removing from the code base before you can apply these changes. I won't approve this PR for now, until I've sorted the Lambda removal out.

TomJKing (approved):
Looks good. Thanks

You will need to merge the latest changes to this branch.

Also I would hold off merging/applying these changes until a production Terraform apply has been done. It's deleting a lot of resources so causes quite a bit of noise so best to until production is up to date with the current changes. Helps avoid any complications.

PR #398, opened 2023-11-17
Remove support for TRE architecture

TomJKing:
Requires PR: https://github.com/nationalarchives/tdr-terraform-modules/pull/270

PR #923, opened 2026-06-10
Tdrd 1547 review and fix security group descriptions

EnyoHA:
All was security groups in the environments repo checked. Changed descriptions for redis, db_migration and eco_tasks groups. Left lb and end_point_security_group the same.

Checked nested repos and descriptions are accurate.

Readied bastion role as had removed from local repo.

### DIFF
--- a/modules/consignment-api/security.tf
+++ b/modules/consignment-api/security.tf
@@ -28,7 +28,7 @@ resource "aws_security_group" "lb" {
 # Traffic to the ECS cluster should only come from the application load balancer
 resource "aws_security_group" "ecs_tasks" {
   name        = "${var.app_name}-ecs-tasks-security-group-${var.environment}"
-  description = "Allow inbound access to the ECS from the keycloak load balancer only"
+  description = "Allow inbound access to the ECS from the consignment API load balancer only"
   vpc_id      = var.vpc_id
 
   ingress {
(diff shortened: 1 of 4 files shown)


modules/consignment-api/security.tf
```diff
@@ -28,7 +28,7 @@ resource "aws_security_group" "lb" {
 # Traffic to the ECS cluster should only come from the application load balancer
 resource "aws_security_group" "ecs_tasks" {
   name        = "${var.app_name}-ecs-tasks-security-group-${var.environment}"
-  description = "Allow inbound access from the keycloak load balancer only"
```
TomJKing, line 31: This traffic from the consignment api load balancer, so the description should be: `... access to the ECS from the consignment api load balancer only"`

TomJKing (commented):
Thanks. Just one minor comment.

Also if you're going to deploy this with the version of the `tdr-terraform-modules` with PR https://github.com/nationalarchives/tdr-terraform-modules/pull/398 merged, then should fix up this PR so its valid (line 568 needs to be deleted).

TomJKing:
@EnyoHA Should park this ticket for now. As changing the SG description means that a new sg will be created and the old one deleted. This could be very problematic. Could you close this PR. Thanks

